PR #59, opened 2021-09-23
Lucas/dev

lucashervier:
This PR main goal:

# Actually handle several callable for BB models 

ffa0a7b is focus on that. Work is done to handle `sklearn.base.BaseEstimator`, `tf.lite.Interpreter` and `xgboost.XGBModel` models. If the model instance is not one of them neither a `Union[tf.keras.Model, tf.keras.layers.Layer, tf.Module`] then we only require model to be callable on `np.ndarray` and to return a either a `np.ndarray`or a `tf.Tensor` making it possible to use wrapper around your model. Tests are added in `test_callable.py` for commom models. More exhaustive tests are provided in a [notebook](https://colab.research.google.com/drive/1W2SYrEsqiOB-MrUXUTP04kNnUXORpXyL?authuser=1).

# Others

f5670cb updates the `README.md`, `TUTORIAL.md` and `docs/index.md` files


README.md
```diff
@@ -235,7 +291,7 @@ This project received funding from the French ”Investing for the Future – PI
 
 ## 👨‍🎓 Creator
 
-This library was started as a side-project by [Thomas FEL](https://twitter.com/Napoolar) who is doing a thesis on explainability.
+This library was started as a side-project by [Thomas FEL](https://twitter.com/Napoolar) who is doing a thesis on explainability. He then received help from some members of the <a href="https://www.deel.ai/"> DEEL </a> team to enhance the library.
```
fel-thomas, line 294: Don't hesitate to put your name on it ;)

README.md
```diff
@@ -205,6 +208,59 @@ All the attributions method presented below handle both **Classification** and *
...
+The model can be something else than a `tf.keras.Model` if it respects one of the following condition:
+- `model(inputs: np.ndarray)` return a `tf.Tensor` of shape $(N, L)$ where $N$ is the number of samples and $L$ the number of targets
+- The model has a `scikit-learn` API and has a `predict_proba` function
+- The model is a `xgboost.XGBModel` from the [XGBoost python library](https://xgboost.readthedocs.io/en/latest/python/python_intro.html)
+- The model is a [TF Lite model](https://www.tensorflow.org/api_docs/python/tf/lite). Note this feature is experimental.
+
+On the other hand, a PyTorch model can be used with method having Callable as type of model. In order to makes it work you should write a
+wrapper as follow:
+
+```python
```
fel-thomas, line 222: I don't know if we need to be that specific in the readme... Like this code block would better fit in a documentation page that we can link here no ? I would keep the list of 4 element and put a link on a doc page ?
What do you think ?

docs/index.md
```diff
@@ -96,9 +102,10 @@ from xplique.metrics import Deletion
 # ...
 
 explainer = GradCAM(model)
+explanations = explainer(inputs, labels)
 metric = Deletion(model, inputs, labels)
 
-score_grad_cam = metric(explainer)
+score_grad_cam = metric(explanations)
```
fel-thomas, line 108: thanks ! ;)

docs/index.md
```diff
@@ -197,6 +208,59 @@ All the attributions method presented below handle both **Classification** and *
   <i>methods with TF need a Tensorflow model. </i>
 </div>
 
+## 📞 Callable
```
fel-thomas, line 211: Same here :)

xplique/attributions/base.py
```diff
@@ -53,8 +56,16 @@ def __init__(self, model: Callable, batch_size: Optional[int] = 64):
             if model_key not in BlackBoxExplainer._cache_models:
                 BlackBoxExplainer._cache_models[model_key] = model
             self.model = BlackBoxExplainer._cache_models[model_key]
+            self.inference_function = predictions_one_hot
+            self.batch_inference_function = batch_predictions_one_hot
+        elif isinstance(model, (tf.Module, tf.keras.layers.Layer)):
```
fel-thomas, line 61: Ok so Sequential Keras.Model and stuff inherit from tf.keras.layers.Layer ? :)

fel-thomas (commented):
Just the remark for the code / text about the wrapper (I suggest we create a page in the documentation instead of surcharging the readme, already too long).

Else LGTM, awesome PR as always ;)

xplique/attributions/base.py
lucashervier: Sequential inherit from tf.keras.Model so it goes into the first if! tf.keras.layers.Layer and tf.Module are a little bit different. I believe though that tf.keras.model inherit from tf.keras.Model so it is import to use the elif after the first one!

README.md
lucashervier: Yes, I put into the docs!

README.md
lucashervier: Well let's go!

fel-thomas (approved):
LGTM :) !

PR #62, opened 2021-10-04
Thomas/wip Feature Viz & Asset enhancement

fel-thomas:
1d4c2bffd5a83b2755fa22d0818287585c03eac0 Apply range value for feature viz allowing to target model in various input range (e.g [-1, 1], [-255, 255]...)

1d4c2bffd5a83b2755fa22d0818287585c03eac0 I tried to make the logo compatible on dark/light mode, nothing extraordinary but I think it's better, what do you think @lucashervier ? ^^


xplique/features_visualizations/preconditioning.py
```diff
@@ -75,6 +75,8 @@ def to_valid_rgb(images: tf.Tensor,
     # rescale according to value range
     images = images - tf.reduce_min(images, (1, 2, 3), keepdims=True)
     images = images / tf.reduce_max(images, (1, 2, 3), keepdims=True)
+    images *= values_range[1] - values_range[0]
```
thib-s, line 78: silent error when value_range is given in wrong order eg (1,-1) instead of (-1, 1)

thib-s (approved):

lucashervier (approved):
Yes it is better! Maybe, not today but we can think about a second logo especially design for dark mode ?

thib-s (approved):

PR #70, opened 2021-11-23
minor: Causal fidelity metrics: Harmonization between timeseries and image - detailed evaluate

AntoninPoche:
# minor: Causal fidelity metrics: Harmonization between timeseries and image - detailed evaluate

This pull request is divided in 3 contributions :

- Harmonization between:
  - CausalFidelity and CausalFidelityTS edcb5c4
  - Adapt testing to harmonization 03d3849
- Detailed evaluate for:
  - CausalFidelity e059af8
  - CausalFidelityTS 4351630
- Remarks in the docs for some metrics: 
  - Insertion and Deletion 4e177cc
  - InsertionTS and DeletionTS af217c7

and a correction for the new version of pylint to work: 22aebd4 (we should use f-string).

### Harmonization

The previously developed causal fidelity metrics for timeseries had several differences in their behavior compared to the initial Insertion and Deletion, thus an harmonization was necessary.

### Detailed evaluate

Furthermore, through discussion, the necessity of accessing all values computed for insertion and deletion was necessary. The idea thus was to add a method (detailed_evaluate) with returns a dictionnary of such values. This method will then be called by the initial evaluate method that will return the auc.

The output of the detailed evaluate method can be used to draw score evolution curves, where the score depend on the number of features perturbed.

### Interpretation and remarks in the documentation

There are remarks about:

- The score interpretation
- How max_percentage_perturbed can influence the score
- The fact that evaluating Insertion and Deletion on the absolute value of explanations may not provide relevant scores.

### DIFF
--- a/docs/api/deletion.md
+++ b/docs/api/deletion.md
@@ -10,7 +10,29 @@ the important pixels.
     
     -- <cite>[RISE: Randomized Input Sampling for Explanation of Black-box Models (2018)](https://arxiv.org/abs/1806.07421)</cite>[^1]
 
-The better the method, the smaller the score.
+
+## Score interpretation
+
+If explanations are accurate, the score will quickly fall from the score on non-perturbed input to the score of a random predictor.
+  Thus, in this case, a lower score represent a more accurate explanation.
+
+
+## Remarks
+
+This metric only evaluate the order of importance between features.
+
+The parameters metric, steps and max_percentage_perturbed may drastically change the score :
+
+- For inputs with many features, increasing the number of steps will allow you to capture more efficiently the difference between attributions methods.
+
+- The order of importance of features with low importance may not matter, hence, decreasing the max_percentage_perturbed,
+may make the score more relevant.
+  
+Sometimes, attributions methods also returns negative attributions,
+for those methods, do not take the absolute value before computing insertion and deletion metrics.
+Otherwise, negative attributions may have higher absolute values, and the order of importance between features will change.
+Therefore, take those previous remarks into account to get a relevant score.
+
 
 ## Example
 
--- a/docs/api/deletion_ts.md
+++ b/docs/api/deletion_ts.md
@@ -6,7 +6,35 @@ This metric computes the capacity of the model to make predictions while perturb
 Specific explanation metrics for time series are necessary because time series and images have different shapes (number of dimensions) and perturbations should be applied differently to them.
 As the insertion and deletion metrics use input perturbation to be computed, creating new metrics for time series is natural[^2].
 
-The better the method, the smaller the score.
+
+## Score interpretation
+
+The interpretation of the score depends on the score metric you are using to evaluate your model.
+- For metrics where the score increases with the performance of the model (such as accuracy).
+If explanations are accurate, the score will quickly fall from the score on non-perturbed input to the score of a random predictor.
+  Thus, in this case, a lower score represent a more accurate explanation.
+  
+- For metrics where the score decreases with the performance of the model (such as losses). 
+If explanations are accurate, the score will quickly rise.
+  Thus, in this case, a higher score represent a more accurate explanation.
+
+
+## Remarks
+
+This metric only evaluate the order of importance between features.
+
+The parameters metric, steps and max_percentage_perturbed may drastically change the score :
+
+- For inputs with many features, increasing the number of steps will allow you to capture more efficiently the difference between attributions methods.
+
+- The order of importance of features with low importance may not matter, hence, decreasing the max_percentage_perturbed,
+may make the score more relevant.
+  
+Sometimes, attributions methods also returns negative attributions,
+for those methods, do not take the absolute value before computing insertion and deletion metrics.
+Otherwise, negative attributions may have higher absolute values, and the order of importance between features will change.
+Therefore, take those previous remarks into account to get a relevant score.
+
 
 ## Example
 
@@ -25,5 +53,5 @@ score = metric.evaluate(explanations)
 
 {{xplique.metrics.DeletionTS}}
 
-[^1]: [RISE: Randomized Input Sampling for Explanation of Black-box Models (2018)](https://arxiv.org/abs/1806.07421)
-[^2]: [Towards a Rigorous Evaluation of XAI Methods on Time Series (2019)](https://arxiv.org/abs/1909.07082)
+[^1]:[RISE: Randomized Input Sampling for Explanation of Black-box Models (2018)](https://arxiv.org/abs/1806.07421)
+[^2]:[Towards a Rigorous Evaluation of XAI Methods on Time Series (2019)](https://arxiv.org/abs/1909.07082)
--- a/docs/api/insertion.md
+++ b/docs/api/insertion.md
@@ -10,7 +10,29 @@ The Insertion Fidelity metric measures how well a saliency-map–based explanati
     
     -- <cite>[RISE: Randomized Input Sampling for Explanation of Black-box Models (2018)](https://arxiv.org/abs/1806.07421)</cite>[^1]
 
-The better the method, the higher the score.
+
+## Score interpretation
+
+If explanations are accurate, the score will quickly rise to the score on non-perturbed input.
+  Thus, in this case, a higher score represent a more accurate explanation.
+
+
+## Remarks
+
+This metric only evaluate the order of importance between features.
+
+The parameters metric, steps and max_percentage_perturbed may drastically change the score :
+
+- For inputs with many features, increasing the number of steps will allow you to capture more efficiently the difference between attributions methods.
+
+- The order of importance of features with low importance may not matter, hence, decreasing the max_percentage_perturbed,
+may make the score more relevant.
+  
+Sometimes, attributions methods also returns negative attributions,
+for those methods, do not take the absolute value before computing insertion and deletion metrics.
+Otherwise, negative attributions may have higher absolute values, and the order of importance between features will change.
+Therefore, take those previous remarks into account to get a relevant score.
+
 
 ## Example
 
--- a/docs/api/insertion_ts.md
+++ b/docs/api/insertion_ts.md
@@ -3,7 +3,35 @@
 The Time Series Insertion Fidelity metric measures the faithfulness of explanations on Time Series predictions[^2].
 This metric computes the capacity of the model to make predictions while only the most important features are not perturbed[^1].
 
-The better the method, the higher the score.
+
+## Score interpretation
+
+The interpretation of the score depends on the score metric you are using to evaluate your model.
+- For metrics where the score increases with the performance of the model (such as accuracy).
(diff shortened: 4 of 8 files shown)


fel-thomas:
You're trying to re-commit commits from last month, can you clean your PR with just the new commits pls ? ;)

docs/api/deletion.md
```diff
@@ -10,7 +10,29 @@ the important pixels.
...
+
+- For inputs with many features, increasing the number of steps will allow you to capture more efficiently the difference between attributions methods.
+
+- The order of importance of features with low importance may not matter, hence, decreasing the max_percentage_perturbed,
+may make the score more relevant.
+  
+Sometimes, attributions methods also returns negative attributions,
+for those methods, do not take the absolute value before computing insertion and deletion metrics.
+Otherwise, negative attributions may have higher absolute values, and the order of importance between features will change.
+Therefore, the score may not represent the pertinence of the explanation.
```
lucashervier, line 34: It is a little bit negative, it sounds like it is useless to use this score which is not true. I rather said something like:
"Therefore, take those previous remarks into account to get a relevant score"

docs/api/deletion_ts.md
```diff
@@ -6,7 +6,35 @@ This metric computes the capacity of the model to make predictions while perturb
 Specific explanation metrics for time series are necessary because time series and images have different shapes (number of dimensions) and perturbations should be applied differently to them.
 As the insertion and deletion metrics use input perturbation to be computed, creating new metrics for time series is natural[^2].
 
-The better the method, the smaller the score.
+
+## Score interpretation
+
+The interpretation of the score depends on the score metric you are using to evaluate your model.
+- For metrics where the score augment with the performance of the model (such as accuracy).
```
lucashervier, line 13: augment -> increase

docs/api/deletion_ts.md
```diff
@@ -6,7 +6,35 @@ This metric computes the capacity of the model to make predictions while perturb
...
+
+- For inputs with many features, increasing the number of steps will allow you to capture more efficiently the difference between attributions methods.
+
+- The order of importance of features with low importance may not matter, hence, decreasing the max_percentage_perturbed,
+may make the score more relevant.
+  
+Sometimes, attributions methods also returns negative attributions,
+for those methods, do not take the absolute value before computing insertion and deletion metrics.
+Otherwise, negative attributions may have higher absolute values, and the order of importance between features will change.
+Therefore, the score may not represent the pertinence of the explanation.
```
lucashervier, line 36: Same as previously

docs/api/insertion.md
```diff
@@ -10,7 +10,29 @@ The Insertion Fidelity metric measures how well a saliency-map–based explanati
...
+
+- For inputs with many features, increasing the number of steps will allow you to capture more efficiently the difference between attributions methods.
+
+- The order of importance of features with low importance may not matter, hence, decreasing the max_percentage_perturbed,
+may make the score more relevant.
+  
+Sometimes, attributions methods also returns negative attributions,
+for those methods, do not take the absolute value before computing insertion and deletion metrics.
+Otherwise, negative attributions may have higher absolute values, and the order of importance between features will change.
+Therefore, the score may not represent the pertinence of the explanation.
```
lucashervier, line 34: Idem

docs/api/insertion_ts.md
```diff
@@ -3,7 +3,35 @@
...
+
+- For inputs with many features, increasing the number of steps will allow you to capture more efficiently the difference between attributions methods.
+
+- The order of importance of features with low importance may not matter, hence, decreasing the max_percentage_perturbed,
+may make the score more relevant.
+  
+Sometimes, attributions methods also returns negative attributions,
+for those methods, do not take the absolute value before computing insertion and deletion metrics.
+Otherwise, negative attributions may have higher absolute values, and the order of importance between features will change.
+Therefore, the score may not represent the pertinence of the explanation.
```
lucashervier, line 33: Idem

tests/metrics/test_fidelity.py
```diff
@@ -50,20 +51,24 @@ def test_perturbation_metrics():
...
-                )(explanations)
-                score_deletion = DeletionTS(
-                    model, x, y, metric="loss", baseline_mode=baseline_mode,
-                    steps=step, max_percentage_perturbed=max_percentage_perturbed,
-                )(explanations)
-
-                for score in [score_insertion, score_deletion]:
-                    assert 0.0 < score < 1
+    for step in [-1, 10]:
+        for max_percentage_perturbed in [0.2]:
```
lucashervier, line 55: Why did you remove a max_percentage_pertubed value ? Especially the for loop is useless in this case ^^

xplique/metrics/fidelity.py
```diff
@@ -198,6 +197,35 @@ def evaluate(self,
...
+
+        Parameters
+        ----------
+        explanations
+            Explanation for the inputs, labels to evaluate.
+
+        Returns
+        -------
+        causal_score_dict
+            Dictionary of scores obtain for different perturbations
```
lucashervier, line 224: While reading this it is very unclear.. Is it for different pertubations ? Or is it the same pertubation with an incremental process ?

xplique/metrics/fidelity.py
```diff
@@ -383,16 +408,11 @@ def __init__(self,
...
         assert 0 < max_percentage_perturbed <= 1, \
             "max_percentage_perturbed should be between 0 and 1"
-        max_nb_perturbation = floor(max_percentage_perturbed*self.nb_features)
+        self.max_percentage_perturbed = max_percentage_perturbed
 
         if steps == -1:
-            steps = max_nb_perturbation
-        assert 1 < steps <= max_nb_perturbation, \
-            f"The number of steps must be between 2 and {max_nb_perturbation} (or '-1')."
+            steps = self.max_percentage_perturbed * self.nb_features
```
lucashervier, line 414: I think steps should be an int. Thus there is a floor or ceil function missing I guess

xplique/metrics/fidelity.py
```diff
@@ -407,8 +427,36 @@ def evaluate(self,
         Returns
         -------
         causal_score
-            Metric score, area over the deletion (lower is better) or insertion (higher is
-            better) curve.
+            Metric score (for interpretation, see score interpretation in the documentation).
```
lucashervier, line 430: It still would be nice to have a summary of the documentation for interpretation

xplique/metrics/fidelity.py
```diff
@@ -407,8 +427,36 @@ def evaluate(self,
...
+
+        Parameters
+        ----------
+        explanations
+            Explanation for the inputs, labels to evaluate.
+
+        Returns
+        -------
+        causal_score_dict
+            Dictionary of scores obtain for different perturbations
```
lucashervier, line 456: What motivates the choice of a dictionnary rather than a list btw ? :)

xplique/metrics/fidelity.py
```diff
@@ -436,9 +484,9 @@ def evaluate(self,
 
         baselines_flatten = baselines.reshape(self.inputs_flatten.shape)
 
-        steps = np.linspace(1, self.max_nb_perturbation, self.steps, dtype=np.int32)
+        steps = np.linspace(0, self.nb_features * self.max_percentage_perturbed, self.steps+1,
```
lucashervier, line 487: Seeing this, I think it is better to define self.max_nb_pertubations as it was previously made!

xplique/metrics/fidelity.py
lucashervier: Why don't you keep max_nb_pertubation ?

docs/api/deletion.md
AntoninPoche: That is a good idea

tests/metrics/test_fidelity.py
AntoninPoche: I had too much values, thus the test was too long and I though that if everyone does that, the testing may last for days in the end.
Thus I will remove the loop

xplique/metrics/fidelity.py
AntoninPoche: It is the same perturbations, the computation is just divided so that an user can have access to all computed values. It is useful to visualize insertion and deletion curves.
Further precisions in the documentation will be added

lucashervier (changes requested):
Most of my remarks question the relevance of some change to get more complete metrics. The idea is good but there is a few change to make I believe or to discuss at least. Otherwise, well done :+1: :tada:

xplique/metrics/fidelity.py
AntoninPoche: The idea was to limit the number of attribute of the class, we discussed it with Thomas, do you think it is not justified ?

xplique/metrics/fidelity.py
AntoninPoche: As aforementioned, this method main use is to allow visualization of insertion and deletion curves.
Thus keeping the number of features pertubed as keys of the dictionnary allow an easier implementation of the visualization and a more pertinent number of perturbation axis.

docs/api/deletion.md
```diff
@@ -10,7 +10,29 @@ the important pixels.
     
     -- <cite>[RISE: Randomized Input Sampling for Explanation of Black-box Models (2018)](https://arxiv.org/abs/1806.07421)</cite>[^1]
 
-The better the method, the smaller the score.
+
+## Score interpretation
+
+If explanations are accurate, the score will quickly fall from the score on non-perturbed input to the score of a random predictor.
```
fel-thomas, line 16: Awesome ;) !

docs/api/deletion_ts.md
```diff
@@ -6,7 +6,35 @@ This metric computes the capacity of the model to make predictions while perturb
 Specific explanation metrics for time series are necessary because time series and images have different shapes (number of dimensions) and perturbations should be applied differently to them.
 As the insertion and deletion metrics use input perturbation to be computed, creating new metrics for time series is natural[^2].
 
-The better the method, the smaller the score.
+
+## Score interpretation
+
+The interpretation of the score depends on the score metric you are using to evaluate your model.
```
fel-thomas, line 12: I like the warning for the TS score, good job

xplique/metrics/fidelity.py
```diff
@@ -198,6 +201,41 @@ def evaluate(self,
...
+        scores_dict = self.detailed_evaluate(explanations)
+
+        # compute auc using trapezoidal rule (the steps are equally distributed)
+        np_scores = np.array(list(scores_dict.values()))
+        auc = np.mean(np_scores[:-1] + np_scores[1:]) * 0.5
+
+        return auc
+
+    def detailed_evaluate(self,
+                          explanations: Union[tf.Tensor, np.ndarray]) -> Dict[int, float]:
```
fel-thomas, line 213: :+1:

fel-thomas (approved):
Cool PR, LGTM ! ;)

lucashervier (approved):
Thank you for the modification!

PR #75, opened 2021-12-09
Precommit conf

justinplakoo:


fel-thomas:
I propose that we stay with pylint for the moment and that we keep our commit convention as indicated in the [Contributing](https://github.com/deel-ai/xplique/blob/master/CONTRIBUTING.md). For the rest -- governance, pre-commit hook, re-ordering -- I agree, great idea ! ;)

GOVERNANCE.md
```diff
@@ -0,0 +1,46 @@
+# GOVERNANCE
+
+`Xplique` is developped as part of the Artificial and Natural Intelligence Toulouse Institute (DEEL/ANITI) program.
```
lucashervier, line 3: This file is good but it sounds a little bit "strict" to me. I think it would be nice to specify somewhere that it is really free to use/exploit for anyone's work and will ?

LICENSES/headers/MIT-Clause.txt
```diff
@@ -0,0 +1,22 @@
+-*- coding: utf-8 -*-
+Copyright IRT Antoine de Saint Exupéry et Université Paul Sabatier Toulouse III - All
```
lucashervier, line 2: It has to be checked but for me you cannot be "All rights reserved" and MIT licence. The copyright is legit though

setup.py
```diff
@@ -1,4 +1,27 @@
-from setuptools import setup, find_packages
+# -*- coding: utf-8 -*-
+# Copyright IRT Antoine de Saint Exupéry et Université Paul Sabatier Toulouse III - All
```
lucashervier, line 2: As said in the MIT licence, I believe that All rights reserved explicitly means that you cannot use it in the context of the MIT licence

tests/__init__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright IRT Antoine de Saint Exupéry et Université Paul Sabatier Toulouse III - All
```
lucashervier, line 2: idem

tests/attributions/__init__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright IRT Antoine de Saint Exupéry et Université Paul Sabatier Toulouse III - All
+# rights reserved. DEEL is a research program operated by IVADO, IRT Saint Exupéry,
```
lucashervier, line 3: idem

tests/features_visualizations/test_regularizers.py
```diff
@@ -18,9 +42,9 @@ def test_lp_norm():
 
 
 def test_tv():
-    vec = tf.cast([[[0., 0., 0.],
-                    [0., 2., 0.],
-                    [0., 0., 1.]]], tf.float32)[:, :, :, tf.newaxis]
+    vec = tf.cast([[[0.0, 0.0, 0.0], [0.0, 2.0, 0.0], [0.0, 0.0, 1.0]]], tf.float32)[
```
lucashervier, line 45: Even though it is better from a syntaxic point of view I think it is harder to know what this is doing at a glance with this format

setup.py
justinplakoo: "When you do some work, the default copyright position is all rights reserved. Nobody can do anything with your work without your express permission, or a license to do so.

The MIT license is your license to other people, to make it clear what they can do. However, any rights not explicitly granted by the license are still reserved by you.

Since the default position is all rights reserved, it's not strictly necessary to add that to a copyright notice - but it does make it clearer, and can help avoid scenarios where someone uses your work illegally.

By having the all rights reserved notice, and then a license, you're affirming it's your work and your rights, but you allow people to do these specific things with it."

lucashervier (commented):
What a tremendous syntaxic work! Well done!
I just want to highlight the fact that I think it does not make sense to have both a:
"All rights reserved" which means that we are the only one able to share, use and exploit it and to have a MIT licence. If we want something in-between then we have to think of something better fitted. I might be wrong but for me it is opposing terms.

@justinplakoo's answer which drop my concerns:
> "When you do some work, the default copyright position is all rights reserved. Nobody can do anything with your work without your express permission, or a license to do so.
> 
> The MIT license is your license to other people, to make it clear what they can do. However, any rights not explicitly granted by the license are still reserved by you.
> 
> Since the default position is all rights reserved, it's not strictly necessary to add that to a copyright notice - but it does make it clearer, and can help avoid scenarios where someone uses your work illegally.
> 
> By having the all rights reserved notice, and then a license, you're affirming it's your work and your rights, but you allow people to do these specific things with it."

setup.py
lucashervier: Ok seems more clear to me :)

fel-thomas:
:tada:  Merged in #77! ;)

PR #77, opened 2021-12-10
JP & TF: pre-commit

fel-thomas:
## Summary

I'm taking @justinplakoo PR in order to try to detail and split the code so that we can discuss about _pre-commit_.

This pull request is to introduce the _pre-commit_ mechanism. The goal is simple: once installed, it will make minor fixes for you before committing. I added the dependency in the requirement_dev and it will be enough to install it with a pip install.

For more information: [Pre-commit](https://pre-commit.com/).

## Introduction of Pre-commit

In details now, _pre-commit_  are defined as hooks in the `pre-commit-config.yaml` file. The hooks that are introduced for the moment are: 
- trailing whitespace (4b1e357c9197b62ff08c06ffd8f0f31cb3db25ef) which simply remove all the bad whitespace 
- large files (0602a4acfe07de1c372a59793080bad9b4f239db) which prevent us to commit large file by mistake
- ast (b3501646c5c2c51c0249ff8f3707a47b65730050) which check whether files parse as valid python
- docstring first (5487d86a340e6ddec7e41b5b57125b499e296c6a) which check a common error of placing code before the docstring
- merge conflict (af17b3f50c2e7411a9a80f736fd95f37bb4b8bd0) which look for merge conflict and prevent us to commit them
- final newline (4373d0f5b4d18a80dfe8f4a87beaebbc2253efcf) which ensure we have a final newline at the end of each file

Finally, we run those checks once in (f509dacf1acee2cf12770943ab79b2fe03abf9c4). 
:pray:  I apologize in advance for all the files affected. I took the liberty of checking all the files one by one to make sure there were no aberrations but a second pair of eyes on this commit would not be too much.

These hooks are automatically triggered during the commits. In case you want to trigger them by hand, @justinplakoo  has put a specific command in the makefile: `make pc_check`.

## Copyright & Governance

On another note, the governance file made by @justinplakoo  is introduced, the existing MIT license where the copyright was in my name is now on the IRT / DEEL / UPS / ANITI (3e770299583d49afd029124fa52de93d55dd0a85). 

## Proposition

I think we should add the pre-commit in the test CI ? What do you think ? If you agree I can prepare another PR to add it.


AntoninPoche (approved):
There seems to be no aberration in the white space cleaning and final line adding.

lucashervier (approved):
Nothing to declare

justinplakoo (approved):
Everything is good for me

thib-s (approved):

PR #83, opened 2022-01-14
AverageStability metric: ensure noisy masks applied use radius parameter

fel-thomas:
@DavidPetiteau and David V.  identified a problem in the stability metric `AverageStability` : the radius is fixed and the parameter is not used. This RP fixes this problem and introduces a test to avoid this in the future.

### DIFF
--- a/xplique/metrics/stability.py
+++ b/xplique/metrics/stability.py
@@ -30,7 +30,8 @@ class AverageStability(ExplainerMetric):
     batch_size
         Number of samples to explain at once, if None compute all at once.
     radius
-        Radius defining the neighborhood of the inputs with respect to l1 distance.
+        Maximum value of the uniform noise added to the inputs before recalculating their
+        explanations.
     distance
         Distance metric between the explanations.
     nb_samples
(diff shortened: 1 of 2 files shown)


xplique/metrics/stability.py
```diff
@@ -60,8 +61,8 @@ def __init__(self,
             raise ValueError(f"{distance} is not a valid distance.")
 
         # prepare the noisy masks that will be used to generate the neighbors
-        nb_variables = np.prod(inputs.shape[1:])
-        self.noisy_masks = tf.random.uniform((nb_samples, *inputs.shape[1:]), 0, 1.0 / nb_variables)
+        self.noisy_masks = tf.random.uniform((nb_samples, *inputs.shape[1:]), 0,
+                                             self.radius)
```
AntoninPoche, line 65: I think that removing the division by nb_variables does not lead to the expected behavior. The noise norm esperance should be the radius, here it is nb_features\*radius.

AntoninPoche (approved):

DavidPetiteau (approved):

PR #85, opened 2022-01-16
Object detector explainer

dejeanph:
2D bounding boxes and segmentation explanation.

### DIFF
--- a/xplique/metrics/fidelity.py
+++ b/xplique/metrics/fidelity.py
@@ -105,7 +105,8 @@ class MuFidelity(ExplanationMetric):
         """
         explanations = np.array(explanations)
         assert len(explanations) == len(self.inputs), "The number of explanations must be the " \
-                                                      "same as the number of inputs"
+                                            f"same as the number of inputs: {len(explanations)}"\
+                                            f" vs {len(self.inputs)}"
 
         correlations = []
         for inp, label, phi, base in zip(self.inputs, self.targets, explanations,
@@ -178,7 +179,7 @@ class CausalFidelity(ExplanationMetric):
         self.inputs_flatten = inputs.reshape((len(inputs), self.nb_features, inputs.shape[-1]))
 
         assert 0.0 < max_percentage_perturbed <= 1.0, "`max_percentage_perturbed` must be" \
-                                                      "in ]O, 1]."
+                                                      "in ]0, 1]."
         self.max_nb_perturbed = tf.math.floor(self.nb_features * max_percentage_perturbed)
 
         if steps == -1:
@@ -238,7 +239,9 @@ class CausalFidelity(ExplanationMetric):
         """
         explanations = np.array(explanations)
         assert len(explanations) == len(self.inputs), "The number of explanations must be the " \
-                                                      "same as the number of inputs"
+                                            f"same as the number of inputs: {len(explanations)}"\
+                                            f"vs {len(self.inputs)}"
+
         # the reference does not specify how to manage the channels of the explanations
         if len(explanations.shape) == 4:
             explanations = np.mean(explanations, -1)
--- a/tests/attributions/test_object_detector.py
+++ b/tests/attributions/test_object_detector.py
@@ -1,0 +1,100 @@
+import numpy as np
+import tensorflow as tf
+
+from xplique.attributions import BoundingBoxesExplainer, Rise
+from xplique.attributions.object_detector import SegmentationIouCalculator, BoxIouCalculator, \
+    ImageObjectDetectorExplainer, ImageObjectDetectorScoreCalculator, IObjectFormater
+
+from ..utils import almost_equal, generate_model, generate_data
+
+
+def test_iou_mask():
+    """Assert the Mask IoU calculation is ok"""
+    dtype = np.float32
+
+    m1 = np.array([
+        [10, 20, 30, 40]
+    ], dtype=dtype)
+    m2 = np.array([
+        [15, 20, 30, 40]
+    ], dtype=dtype)
+    m3 = np.array([
+        [0, 20, 10, 40]
+    ], dtype=dtype)
+    m4 = np.array([
+        [0, 0, 100, 100]
+    ], dtype=dtype)
+
+    iou_calculator = BoxIouCalculator()
+
+    assert almost_equal(iou_calculator.intersect(m1, m2), 300.0 / 400.0)
+    assert almost_equal(iou_calculator.intersect(m1, m3), 0.0)
+    assert almost_equal(iou_calculator.intersect(m3, m2), 0.0)
+    assert almost_equal(iou_calculator.intersect(m1, m4), 400.0 / 10_000)
+    assert almost_equal(iou_calculator.intersect(m2, m4), 300.0 / 10_000)
+    assert almost_equal(iou_calculator.intersect(m3, m4), 200.0 / 10_000)
+
+
+def test_iou_segmentation():
+    """Assert the segmentation IoU computation is ok"""
+
+    m1 = np.array([
+        [0, 0],
+        [1, 1],
+    ])[None, :, :]
+    m2 = np.array([
+        [1, 1],
+        [0, 0],
+    ])[None, :, :]
+    m3 = np.array([
+        [1, 0],
+        [1, 0],
+    ])[None, :, :]
+
+    iou_calculator = SegmentationIouCalculator()
+
+    assert almost_equal(iou_calculator.intersect(m1, m2), 0.0)
+    assert almost_equal(iou_calculator.intersect(m1, m3), 1.0/3.0)
+    assert almost_equal(iou_calculator.intersect(m3, m2), 1.0/3.0)
+    assert almost_equal(iou_calculator.intersect(m1, m1), 1.0)
+    assert almost_equal(iou_calculator.intersect(m2, m2), 1.0)
+
+
+def test_image_object_detector():
+    """Assert input shape returned is correct"""
+    input_shape = (8, 8, 1)
+    nb_labels = 2
+    x, y = generate_data(input_shape, nb_labels, nb_labels)
+    model = generate_model(input_shape, nb_labels)
+
+    method = Rise(model, nb_samples=10)
+
+    obj_ref = tf.cast([
+                [0, 0, 100, 100, 0.9, 1.0, 0.0],
+            ], tf.float32)
+
+    class BBoxFormater(IObjectFormater):
+
+        def format_objects(self, predictions):
+            if np.all(predictions.numpy() == obj_ref.numpy()):
+                return obj_ref[:4], obj_ref[4:5], obj_ref[5:]
+
+
+            bboxes = tf.cast([
+                [0, 10, 20, 30],
+                [0, 0, 100, 100],
+            ], tf.float32)
+
+            proba = tf.cast([0.9, 0.1], tf.float32)
+
+            classif = tf.cast([[1.0, 0.0], [0.0, 1.0]], tf.float32)
+
+            return bboxes, proba, classif
+
+    formater = BBoxFormater()
+    explainer = ImageObjectDetectorExplainer(method, formater, BoxIouCalculator())
+
+    phis = explainer(x, obj_ref)
+
+    assert phis.shape == (1, input_shape[0], input_shape[1])
+
--- a/xplique/attributions/__init__.py
+++ b/xplique/attributions/__init__.py
@@ -16,3 +16,4 @@ from .deconvnet import DeconvNet
 from .grad_cam_pp import GradCAMPP
 from .lime import Lime
 from .kernel_shap import KernelShap
+from .object_detector import BoundingBoxesExplainer
--- a/xplique/attributions/object_detector.py
+++ b/xplique/attributions/object_detector.py
@@ -1,0 +1,284 @@
+"""
+Module related to Object detector method
+"""
+
+from typing import Iterable, Tuple, Union, Optional
+import abc
+
+import tensorflow as tf
+import numpy as np
+
+from xplique.attributions.base import BlackBoxExplainer
+from xplique.commons.tf_operations import inference_batching
+
+
+class IouCalculator:
+    """
+    Used to compute the Intersection Over Union (IOU).
+    """
+    @abc.abstractmethod
+    def intersect(self, objects_a: tf.Tensor, objects_b: tf.Tensor) -> tf.Tensor:
+        """
+        Compute the intersection between two batched objects (e.g boxes, segmentation masks...)
+
+        Parameters
+        ----------
+        objects_a
+            First batch of objects to compare.
+        objects_b
+            Second batch of objects to compare.
+
+        Returns
+        -------
+        score
+            Real value between [0,1] corresponding to the intersection of the 2 objects.
+        """
+        raise NotImplementedError()
+
+
+class SegmentationIouCalculator(IouCalculator):
+    """
+    Compute segmentation masks IOU.
+    """
+    def intersect(self, masks_a: tf.Tensor, masks_b: tf.Tensor) -> tf.Tensor:
+        """
+        Compute the intersection between two batched segmentation masks.
+        Each segmentation is a boolean mask on the whole image
+
+        Parameters
+        ----------
+        masks_a
+            First batch of segmentation masks.
+        masks_b
+            Second batch of segmentation masks.
+
+        Returns
+        -------
+        iou_score
+            The IOU score between the first and second batch of masks.
+        """
+        # pylint: disable=W0221,W0237
+
+        axis = np.arange(1, tf.rank(masks_a))
+
+        inter_area = tf.reduce_sum(tf.cast(tf.logical_and(masks_a, masks_b),
+                                    dtype=tf.float32), axis=axis)
+        union_area = tf.reduce_sum(tf.cast(tf.logical_or(masks_a, masks_b),
+                                    dtype=tf.float32), axis=axis)
+
+        iou_score = inter_area / tf.maximum(union_area, 1.0)
+
+        return iou_score
+
+
+
+class BoxIouCalculator(IouCalculator):
+    """
+    Used to compute the Bounding Box IOU.
+    """
+    EPSILON = tf.constant(1e-4)
+
+    def intersect(self, boxes_a: tf.Tensor, boxes_b: tf.Tensor) -> tf.Tensor:
+        """
+        Compute the intersection between two batched bounding boxes.
+        Each bounding box is defined by (x1, y1, x2, y2) respectively (left, bottom, right, top).
+
+        Parameters
+        ----------
+        boxes_a
+            First batch of bounding boxes.
+        boxes_b
+            Second batch of bounding boxes.
+
+        Returns
+        -------
+        iou_score
+           The IOU score between the first and second batch of bounding boxes.
+        """
+        # pylint: disable=W0221,W0237
+
+        # determine the intersection rectangle
+        left   = tf.maximum(boxes_a[..., 0], boxes_b[..., 0])
+        bottom = tf.maximum(boxes_a[..., 1], boxes_b[..., 1])
+        right  = tf.minimum(boxes_a[..., 2], boxes_b[..., 2])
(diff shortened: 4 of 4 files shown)


xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
+"""
```
lucashervier, line 1: To follow the current linting convention (and match the other files) it would look like this:
```
"""
Module related to Object detector method
"""

from typing import Iterable, Tuple, Union, Optional
import abc

import tensorflow as tf
import numpy as np

from xplique.attributions.base import BlackBoxExplainer
from xplique.commons.tf_operations import inference_batching
```

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+class IIouCalculator:
+    """
+    Used to compute the IOU.
+    """
+    @abc.abstractmethod
+    def intersect(self, obj1: tf.Tensor, obj2: tf.Tensor) -> tf.Tensor:
+        """
+        Compute the intersection between two batched objects
+        An object could be box, segmentation mask...
+        :param obj1: object1
```
lucashervier, line 21: The documentation being generated with mkdocs, can you please comply to the documentation convention in the other files ? 
```
 """
 Compute the intersection between two batched objects
 An object could be a box, segmentation mask...
 
Parameters
------------
obj1
       object number 1. Can be a box, a segmentation masks, etc..
obj2
       object number 2.  Can be a box, a segmentation masks, etc..

Returns
--------
score
        A score between [0,1] corresponding to the intersection of the 2 objects.
"""
```

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        raise NotImplementedError()
+
+
+class SegmentationIouCalculator(IIouCalculator):
+    """
+    Used to compute the Mask IOU.
+    """
+    def intersect(self, mask1: tf.Tensor, mask2: tf.Tensor) -> tf.Tensor:
+        """
```
lucashervier, line 33: Idem

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+
+class BoxIouCalculator(IIouCalculator):
+    """
+    Used to compute the Bounding Box IOU.
+    """
+    EPSILON = tf.constant(1e-4)
+
+    def intersect(self, box1: tf.Tensor, box2: tf.Tensor) -> tf.Tensor:
+        """
+        Compute the intersection between two batched bounding boxes.
```
lucashervier, line 59: Doc :)

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+
+        return inter_area / tf.maximum(union_area, 1.0)
+
+
+
+class BoxIouCalculator(IIouCalculator):
+    """
+    Used to compute the Bounding Box IOU.
+    """
+    EPSILON = tf.constant(1e-4)
```
lucashervier, line 55: Maybe it could be interesting to be able to change this parameter ? It might be interesting to initialize the BoxIouCalculator with this epsilon as a parameter ?

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+"""
+from typing import Iterable, Tuple, Union, Optional
+import abc
+import tensorflow as tf
+import numpy as np
+from xplique.attributions.base import BlackBoxExplainer
+from xplique.commons.tf_operations import inference_batching
+
+
+class IIouCalculator:
```
lucashervier, line 12: Is it on purpose that there is two capital I . Does it make sense to make the following change:
"IIouCalculator -> IouCalculator" ?

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+        # area and dividing it by the sum of prediction + ground-truth
+        # areas - the interesection area
+        iou = inter_area / (box1_area + box2_area - inter_area + BoxIouCalculator.EPSILON)
+        # return the intersection over union value
+        return iou
+
+
+class IObjectFormater:
+    """
+    Generiq class to format the model prediction
```
lucashervier, line 88: Generic*

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    def format_objects(self, predictions) -> Iterable[Tuple[tf.Tensor, tf.Tensor, tf.Tensor]]:
+        """
+        Format the model prediction of a given image to have the prediction of the following format:
```
lucashervier, line 92: Docs :)

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+        objects, proba_detection, one_hots_classifications
+        :param predictions: prediction of the model of a given image
+        :return: list of tuple: "object,proba,classification"
+        """
+        raise NotImplementedError()
+
+
+class ImageObjectDetectorScoreCalculator:
+    """
+    Class to compute batch score
```
lucashervier, line 102: Docs

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    def __init__(self, object_formater: IObjectFormater, iou_calculator: IIouCalculator):
+        self.object_formater = object_formater
+        self.iou_calculator = iou_calculator
+
+    def batch_score(self, model, inputs, targets, batch_size):
+        """
+        Compute batch score
```
lucashervier, line 110: Docs

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+        :param model: the model used for the object detection
+        : inputs: Input samples to be explained.
+        : targets: model outputs
+        : batch_size: Number of masked samples to explain at once, if None process all at once.
+        """
+        return inference_batching(self.score, model, inputs, targets, batch_size)
+
+    def score(self, model, inp, object_ref) -> tf.Tensor:
+        """
+        Compute the matching score between prediction and a given object
```
lucashervier, line 120: Docs

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+                score_values.append(score_value)
+
+        score_values = tf.stack(score_values)
+
+        return score_values
+
+
+class ImageObjectDetectorExplainer(BlackBoxExplainer):
+    """
+    Used to define method as an object detector one
```
lucashervier, line 167: Docs

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+
+    def __init__(self, explainer: BlackBoxExplainer, object_detector_formater: IObjectFormater,
+                 iou_calculator: IIouCalculator):
+        """
+        :param explainer: the black box explainer used to explain the object detector model
```
lucashervier, line 173: Docs

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def explain(self, inputs: Union[tf.data.Dataset, tf.Tensor, np.array],
+                targets: Optional[Union[tf.Tensor, np.array]] = None) -> tf.Tensor:
+        if len(tf.shape(targets)) == 1:
+            targets = tf.expand_dims(targets, axis=0)
+
+        return self.explainer.explain(inputs, targets)
+
+
+class BoundingBoxesExplainer(ImageObjectDetectorExplainer, IObjectFormater):
```
lucashervier, line 193: Well done ;)

lucashervier (changes requested):
Nice work, it will be really helpful! However, can you please change the documentation style to match all the other files ?
In addition, it needs a corresponding test files !
Can you please also edit your PR message and link the corresponding issue ? :)

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: It's just a tiny value in case of division by zero.
(Cf. Rise implementation for example)

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Done.

xplique/attributions/image_object_detector_explainer.py
lucashervier: 👍

lucashervier (approved):
Thanks for the modification! LGTM :)

xplique/metrics/fidelity.py
```diff
@@ -178,7 +179,7 @@ def __init__(self,
         self.inputs_flatten = inputs.reshape((len(inputs), self.nb_features, inputs.shape[-1]))
 
         assert 0.0 < max_percentage_perturbed <= 1.0, "`max_percentage_perturbed` must be" \
-                                                      "in ]O, 1]."
+                                                      "in ]0, 1]."
```
fel-thomas, line 182: I propose we stay consistent with the actual code and stay with (the new) f-string way when we can. :)
https://www.python.org/dev/peps/pep-0498/

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
+"""
+Module related to Object detector method
```
fel-thomas, line 2: Do we need a proper folder to store the bouding box explainers or 1 file is enough ? @dejeanph what do you think (in the future I mean)?

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+                            self.object_formater.format_objects(obj)
+
+                if len(tf.shape(obj_ref)) == 1:
+                    obj_ref = tf.expand_dims(obj_ref, axis=0)
+
+                obj_ref = self.object_formater.format_objects(obj_ref)
+
+                scores = []
+                size = tf.shape(current_boxes)[0]
+                for boxes_ref, proba_ref, class_ref in zip(*obj_ref):
```
fel-thomas, line 202: Not an obligation but It could be cool to have some comments here to understand what' going on. I will try to adress that

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    @abc.abstractmethod
+    def intersect(self, obj1: tf.Tensor, obj2: tf.Tensor) -> tf.Tensor:
+        """
+        Compute the intersection between two batched objects
+        An object could be box, segmentation mask...
+
+        Parameters
+        ----------
+        obj1
```
fel-thomas, line 27: better use pythonic variable name such as obj_a or obj_1 if you want. (it's minor just for the next PRs)

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+        mask2
+            mask of segmentation 2
+
+        Returns
+        -------
+        iou
+            The IOU score between mask1 and mask2
+        """
+        # pylint: disable=W0237
+        axis = np.arange(1, len(tf.shape(mask1)))
```
fel-thomas, line 62: you can use `tf.rank`, but I think you're not the only one to do that in the codebase :)

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+    EPSILON = tf.constant(1e-4)
+
+    def intersect(self, box1: tf.Tensor, box2: tf.Tensor) -> tf.Tensor:
+        """
+        Compute the intersection between two batched bounding boxes.
+        The bounding box is defined by (x1, y1, x2, y2)
+
+        Parameters
+        ----------
+        boxA
```
fel-thomas, line 85: The parameters does not correspond to the signature, moreover you use Maj in the variables, please if we can we should stay pythonic ;)

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+        inter_area = tf.maximum(0, x2 - x1) * tf.maximum(0, y2 - y1)
+        # compute the area of both the prediction and ground-truth
+        # rectangles
+        box1_area = (box1[..., 2] - box1[..., 0]) * (box1[..., 3] - box1[..., 1])
+        box2_area = (box2[..., 2] - box2[..., 0]) * (box2[..., 3] - box2[..., 1])
+        # compute the intersection over union by taking the intersection
+        # area and dividing it by the sum of prediction + ground-truth
+        # areas - the interesection area
+        iou = inter_area / (box1_area + box2_area - inter_area + BoxIouCalculator.EPSILON)
+        # return the intersection over union value
```
fel-thomas, line 112: I think you can remove this comment

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+        objects, proba_detection, one_hots_classifications
+
+        Parameters
+        ----------
+        predictions
+            prediction of the model of a given image
+
+        Returns
+        -------
+        object
```
fel-thomas, line 132: I think your comments is good, we should call this variable bbox no ?

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        # pylint: disable=W0237
+        # pylint: disable=C0103
+        # determine the (x, y)-coordinates of the intersection rectangle
+        x1 = tf.maximum(box1[..., 0], box2[..., 0])
+        y1 = tf.maximum(box1[..., 1], box2[..., 1])
+        x2 = tf.minimum(box1[..., 2], box2[..., 2])
+        y2 = tf.minimum(box1[..., 3], box2[..., 3])
+        # compute the area of intersection rectangle
+        inter_area = tf.maximum(0, x2 - x1) * tf.maximum(0, y2 - y1)
```
fel-thomas, line 103: Why do we need to use `tf.maximum(.)` here ? I don't follow, either we should assert at the top that for each box we have x2 > x1, or we assume it's always true, but I don't like to put max here, It could be a silent error (e.g a user alway get 0 but don't understand why).

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+        # compute the area of both the prediction and ground-truth
+        # rectangles
+        box1_area = (box1[..., 2] - box1[..., 0]) * (box1[..., 3] - box1[..., 1])
+        box2_area = (box2[..., 2] - box2[..., 0]) * (box2[..., 3] - box2[..., 1])
```
fel-thomas, line 107: Also you don't do the max here, so I propose we remove it up there, it doesnt make sense no ? ;)

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+                if len(tf.shape(obj_ref)) == 1:
+                    obj_ref = tf.expand_dims(obj_ref, axis=0)
+
+                obj_ref = self.object_formater.format_objects(obj_ref)
+
+                scores = []
+                size = tf.shape(current_boxes)[0]
+                for boxes_ref, proba_ref, class_ref in zip(*obj_ref):
+                    boxes_ref = tf.repeat(tf.expand_dims(boxes_ref, axis=0), repeats=size, axis=0)
+                    proba_ref = tf.repeat(tf.expand_dims(proba_ref, axis=0), repeats=size, axis=0)
```
fel-thomas, line 204: proba_ref is not used ? why ? ;)

xplique/attributions/image_object_detector_explainer.py
```diff
@@ -0,0 +1,280 @@
...
+        inputs
+            Input samples to be explained.
+        targets
+            model outputs
+        batch_size
+            number of masked samples to explain at once, if None process all at once.
+        """
+        return inference_batching(self.score, model, inputs, targets, batch_size)
+
+    def score(self, model, inp, object_ref) -> tf.Tensor:
```
fel-thomas, line 167: we need the types in the signature

fel-thomas (changes requested):
first pass which seems good. 
I put some corrections, I will try to push a part of my comments to facilitate you the task. **Nevertheless some important things are missing: unit tests.** We will need some unit test to ensure non regression and to be sure everything work as intendend.

Once the corrections are done and the new tests are ok, we will merge it! ;) 

**EDIT**: I tried to adress some of my comment in the last commit, but the tests still need to be done ;)

xplique/metrics/fidelity.py
dejeanph: It was just a comment : O instead of Zero

xplique/attributions/image_object_detector_explainer.py
dejeanph: Each explained bounding box is drawn in a single image. If x bounding boxes are detected, x image results have to be save.
So a folder is usefull. By default, the algorithm can add the id of the bounding box as a suffix to the image filename.

xplique/attributions/image_object_detector_explainer.py
dejeanph: Cf paper. The algorithm implement the process described in the article about D-Rise.

xplique/attributions/image_object_detector_explainer.py
dejeanph: As said in the article, proba_ref is part of the score.
In this case of bounding box problem, proba_ref has the value 1.
We let it for next problems (segmentation, etc.)

xplique/attributions/image_object_detector_explainer.py
fel-thomas: I mean a folder in the attribution folder, instead of `xplique/attributions/bbox_explainer.py` => `xplique/attributions/bbox/explainer.py` ? 
But I agree, let's start with that

fel-thomas (approved):
I added the tests and adressed my concerns. 
I think nevertheless that other tests, should be added, as mine are not as good as I wanted.

PR #88, opened 2022-01-20
Avoid warning using plot_feature_impact

DavidPetiteau:
When you call this function with a lot of functions, you'll have a lot of warnings.


lucashervier:
It appears you do not pass the lynting test. Can you please run pylint locally before pushing ? In addition, it will help you better to find the linting issues

DavidPetiteau:
I pushed locally successfully
what can I do now ?

DavidPetiteau:
I successfully pass the pylint and the pytest tests.
what should I do now ?

PR #97, opened 2022-02-24
Attributions: Introduce Sobol attribution method

fel-thomas:
# Sobol Attribution Method

In this PR, I propose to introduce the Sobol' Attribution Method into Xplique Attribution module. 
This attribution method is based on the following paper (https://arxiv.org/abs/2111.04138), and use code from (https://github.com/fel-thomas/Sobol-Attribution-Method) 😄 

In order to work, this method needs
(1) an estimator of the Sobol' indices 
(2) a sampling tool (QMC, LHS, ...)
(2) an explainer

Before introducing each of these components, I propose to fix some links issues / tests.

- 5bf0f7e3c4c72ea82123094bc50dd11f5400401b, c36e786615769a69c50a199d36d5eb0f4168e2ce fix links in tutorials & colabs links.
- 429e0b5f556887823d7e657254e4b0886a499a26 I propose we move to cv2 for image extrapolation to avoid the tensorflow resizing when possible.
- f18ebf69521463f492678d6e4ded5a3d07de1334 I propose to move to github CI again to relax our machines (and it's quicker).
- 700473e835b57ff53b77a1660f13f0faa10abb55 Relax dtype of attribution to allow either numpy or tf.tensor.

### Now is the hard part. First the (1) Estimators: 
80775aee03d3d3b317b63b8383b0ee286037f97f Introduce some Sobol' total indices estimators. The estimator are based on the following papers:

- *Jansen*, Ref. Jansen, M., Analysis of variance designs for model output (1999), https://www.sciencedirect.com/science/article/abs/pii/S0010465598001544
- *Homma*, Ref. Homma & al., Importance measures in global sensitivity analysis of nonlinear models (1996) https://www.sciencedirect.com/science/article/abs/pii/0951832096000026
- *Janon*,  Ref. Janon & al., Asymptotic normality and efficiency of two Sobol index estimators (2014) https://hal.inria.fr/hal-00665048v2/document
- *Glen & Isaacs*, Ref. Glen & al., Estimating Sobol sensitivity indices using correlations (2012) https://dl.acm.org/doi/abs/10.1016/j.envsoft.2012.03.014
- *Saltelli*, Ref. Satelli & al., Global Sensitivity Analysis. The Primer. https://onlinelibrary.wiley.com/doi/book/10.1002/9780470725184

Note that the estimator recommended by the paper is _Jansen_.

### (2) The Samplers

in a5cbf680b29050a33d83e711bed8f0a82499dcc0, I propose multiple samplers that we can use with the estimators:

- Quasi-Monte Carlo Sobol' Sequence sampler based on scipy `qmc` module. We need Scipy 1.7 to use the `qmc` module, and scipy 1.7 need python >= 3.7 which mean dropping python 3.6 if we add it to the requirements. That's why I propose the first `try ... import` from the lib. The way I managed it: wrap every scipy `qmc` sampler in a specific class that raise appropriate warning. Then I handle this specific case during the tests.
- Latin Hypercube Sampling using Scipy `qmc` module. 
- Halton Quasi-Monte Carlo sequence from Scipy `qmc` module.
- TF Sobol sequence sampler. This sampler is the only one using Tensorflow instead of Scipy.  If scipy is available, I strongly advise to switch to the scipy sampler. 

### (3) The Explainer

bf3d0374891d26f23115bf638c81be780503a331 Introduce the regular Explainer with `@sanitize`, that pass all the tests.
I added the corresponding Documentation page, and propose a test based on the Ishigami function to ensure we get consistent results with all the estimators (which is the case!).

If someone could double-check the test for me it would be cool. ;) 
Ishigami function
```python
Y = sin(X1) + a sin(X2)^2 + b * X3^4 * sin(X1)
X in U([-pi, pi])^3

# analytically we find:
V(Y) = a^2 / 8 + b*pi^4/5 + b^2pi^8/18 + 1/2

V1 = b * pi^4 / 5 + b^2 * pi^8 / 50 + 1/2
V2 = a^2/8
V3 = 0

V12 = 0, V23 = 0, V13 = 8b^2pi^8/225
V13 = 8b^2 pi^8 / 255
V23 = 0

V123 = 0

S1 = V1 / V(Y)
S2 = V2 / V(Y)
S3 = V3 / V(Y)

S13 = V13 / V(Y)

ST1 = S1 + S12 + S13 + S123 = S1 + S13
ST2 = S2 + S12 + S23 + S123 = S2
ST3 = S3 + S13 + S23 + S123 = S13
```

Sorry for the massive PR. If it's too large I can split it.


setup.py
```diff
@@ -10,7 +10,7 @@
     long_description=README,
     long_description_content_type="text/markdown",
     author="Thomas FEL",
-    author_email="thomas.fel@brown.edu",
+    author_email="thomas_fel@brown.edu",
     license="MIT",
     install_requires=['tensorflow>=2.1.0', 'numpy', 'scikit-learn', 'scikit-image',
```
justinplakoo, line 15: In install_requires adding cv2 is missing

AntoninPoche (approved):
I do not have the competences to understand everything that was written around Sobol indexis and estimators, but seeing this code inherit from the code you've done for a paper, I will approve it.

setup.py
fel-thomas: Done ! ;)

PR #98, opened 2022-02-24
Prepare Sobol' Attribution Method

fel-thomas:
_This PR's prepare the introduction of our Sobol Attribution Method._

Few minor bug fixing (d930027768ffd89f75aed5241b17e534b44803c4, 74172fcb08f116c17acc8cf2091647c1842d4c30)
One because the links in the documentation are differents than the one on the readme. 
The second commit is here to remove the `authuser` in the urls of the colab -- users can now access and modify directly as their colabs.

Now the serious stuff: I propose to move the min tensorflow requirements to 2.4. Checking for (2.2, ..., 2.8) is becoming painfull even with the CI, I propose we move forward and stuck with 2.4 for a bit. We should check with the teams but I bet we are all beyond 2.4. ;)  39c7caebabe80ff338eddb3e5cdfdce8c0aeeb6a

To have a good sampling strategy in the Sobol Attribution Method, we need QMC sampling, which mean Scipy >= 1.7 (00ba60d51e0c8268f27af07f809a560e7025e0f5), again, not the newest version but we didn't had any lower requirement on the version.

Finnaly, I propose to introduce `open-cv` in the lib as a lot of people doing ML have already cv2 in their pandas or pip bundle. It will be a huge gain for us if we could use the extrapolation (resize) of CV2 (06f47ff167dcb10b068f08708f343458f81d4b3a).

Once this PR is ok (if you agree with the version requirements), I'll prepare the next one. (#97)


fel-thomas:
After discussion, see directly #97. ;)

PR #99, opened 2022-03-01
Correct wrapper bug and add corresponding testing

### LINKED ISSUE #93: [Bug] [Important] attributions of sklearn wrapped models
---
name: Bug report
about: attributions of sklearn wrapped models
title: "[BUG] - attributions of sklearn wrapped models are incoherent"
labels: ''wrapper", "bug", "attributions"
assignees: ''

---

**Select the modules to which the bug refers:**

- [x] Attributions Methods
- [ ] Feature Visualization
- [ ] Concepts
- [ ] Metrics 
- [ ] Documentation

**Describe the bug**
When a wrapper is used on regression model from sklearn, the obtained attributions are not coherent. (Problem can be larger).

**Screenshots**
wrapper_bug

**Desktop:**
 - OS - windows
 - Python version - 3.7.12
 - Sklearn - 1.0.2
 - Xplique - 0.2.6

**To Reproduce**
```
import numpy as np
from numpy.random import seed, normal
import sklearn
import xplique
from sklearn.linear_model import LinearRegression
from sklearn.model_selection import train_test_split

seed(0)

# dataset parameters
features_coef = np.array([1, -2, -3, 4, 0, 0, 0, 0])
nb_samples = 1000
nb_features = len(features_coef)


# create dataset
dataset = normal(0, 1, (nb_samples, nb_features))
noise = normal(0, 1, nb_samples)
target = dataset.dot(features_coef) + noise

# split dataset
X_train, X_test, y_train, y_test = train_test_split(dataset, target, test_size=0.05, shuffle=True)


# train model
sk_linear = LinearRegression().fit(X_train, y_train)

# Create the wrapper class
class Wrapper():
    # The init method is necessary for every class
    def __init__(self, model):
        self.model = model

    # The call method calls the predict method
    def __call__(self, inputs):
        return self.model.predict(inputs)

# wrap model
sk_model = Wrapper(sk_linear)

# adapt inputs/outputs
inputs_tf = tf.cast(X_test, tf.float32)
targets_tf = tf.ones((X_test.shape[0], 1))

# instanciate explainer
explainer = KernelShap(
    model,
    nb_samples=200,  # 2000
    ref_value=0.0
)

# compute explanations
explanations = abs(explainer(inputs_tf, targets_tf))

print(np.array(explanations).mean(axis=0))

# [4.328179, 4.357149, 4.6055717, 5.554367, 3.5661576, 4.1552, 3.5590754, 4.7494626]
```
To ease the debugging, here is a minimal example notebook (There are several lines for the visualization, but you can jump to the end of the notebook to see the different attributions) : https://colab.research.google.com/drive/1zvt4I9lVpvzg1oWPUNZoFs39w8MPSz_b?usp=sharing

**Expected behavior**
The 4 last attributions values should be close to 0, far inferior to the 4 first.

**Additional context**
_

AntoninPoche:
# Wrapper bug correction and first generic testing

This pull request can be divided in two parts:
- Resolve #93 5499b91
- A new generic testing for the wrapper cecc89f

### The bug correction

The bug reported and described in #93 was corrected as suggested. In `xplique.common.callable_operations.predictions_ont_hot_callable()` a test is effectuated to verify if the target dimension disappeared in the prediction.

### The new test

The issue #93 was created with a minimal example of the bug, however the bug is now fixed. Nevertheless, I figured that I could use this code to build a test for the wrapper and a first generic test.

The test process is the following : 
- A dataset where the output is a linear combinaison of the inputs is created (some inputs are not used and noise is added).
- Two linear regression models are created and trained on the dataset. One with Keras and one with sklearn (Wrapped afterward).
- We verify that the prediction of both models are close (thus weights are close).
- Then we compare method by method that the explanations on both models are close.

Note that only Lime, KernelShap and Occlusion are tested as they are the only method applicable to the LinearRegression model from the Scikit-Learn library.

### DIFF
--- a/tests/generic/test_wrapper.py
+++ b/tests/generic/test_wrapper.py
@@ -1,0 +1,82 @@
+"""
+Test for the wrapper
+Also serve as a generic test
+"""
+
+import numpy as np
+import tensorflow as tf
+
+from xplique.attributions import KernelShap, Lime, Occlusion
+from ..utils import generate_dataset, generate_linear_model, almost_equal
+
+
+def test_wrapper():
+    """
+    Test the wrapping of a sklearn model, by comparing it to a keras model.
+    Both are linear regression model with really close weights.
+    We expect the explanations on those models to be close as well.
+    """
+    nb_samples = 100
+    epsilon = 1.e-4 * nb_samples
+    dataset, _, features_coef = generate_dataset(nb_samples)
+
+    libraries = [
+        "sklearn",
+        "keras",
+    ]
+
+    models = {
+        lib: generate_linear_model(features_coef, lib)
+        for lib in libraries
+    }
+
+    # we want both models to be close enough
+    y_preds = [np.array(model(dataset)).reshape((nb_samples,)) for model in models.values()]
+    assert almost_equal(y_preds[0], y_preds[1], epsilon)
+
+    # Define explainers
+    model_explainers = {}
+    for library, model in models.items():
+        model_explainers[library] = {
+            "KernelShap": KernelShap(
+                model,
+                nb_samples=200,  # 2000
+                ref_value=0.0,
+                batch_size=nb_samples
+            ),
+            "Lime": Lime(
+                model,
+                nb_samples=2000,  # 2000
+                ref_value=0.0,
+                distance_mode="cosine",
+                kernel_width=1.0,
+                batch_size=nb_samples
+            ),
+            "Occlusion": Occlusion(
+                model,
+                patch_size=1,
+                patch_stride=1,
+                occlusion_value=0.0,
+                batch_size=nb_samples
+            ),
+        }
+
+    # cast a subset to tf format
+    inputs_tf = tf.cast(dataset, tf.float32)
+    targets_tf = tf.ones((len(dataset), 1))
+
+    explanations = {}
+
+    # compute explanations for all methods
+    for library, explainers in model_explainers.items():
+        explanations[library] = {}
+        for method, explainer in explainers.items():
+            # Lime is not stable compared, a seed is needed
+            tf.random.set_seed(0)
+            # compute explanation for a given method
+            explanations[library][method] = explainer(inputs_tf, targets_tf)
+
+    for method in explanations[libraries[0]].keys():
+        expl0 = explanations[libraries[0]][method]
+        expl1 = explanations[libraries[1]][method]
+        assert almost_equal(expl0, expl1, epsilon)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,5 +1,7 @@
 import numpy as np
-from tensorflow.keras.models import Sequential
+from sklearn.linear_model import LinearRegression
+import tensorflow as tf
+from tensorflow.keras.models import Sequential, Model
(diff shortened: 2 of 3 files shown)


tests/generic/test_wrapper.py
```diff
@@ -0,0 +1,80 @@
...
+    for library, explainers in model_explainers.items():
+        explanations[library] = {}
+        for method, explainer in explainers.items():
+            # compute explanation for a given method
+            explanations[library][method] = explainer(inputs_tf, targets_tf)
+
+    for method in explanations[libraries[0]].keys():
+        expl0 = explanations[libraries[0]][method]
+        expl1 = explanations[libraries[1]][method]
+        assert mse(expl0, expl1) <= 0.01
```
fel-thomas, line 80: you can use `almost equal` if you want here (but not an obligation).

tests/utils.py
```diff
@@ -50,3 +52,57 @@ def generate_regression_model(features_shape, output_shape=1):
...
+            # The call method calls the predict method
+            def __call__(self, inputs):
+                pred = self.model.predict(inputs)
+                return pred
+
+        sk_model = Wrapper(sk_linear)
+        return sk_model
+
+    elif library == "keras":
+        RANDOM_SEED = 1
```
fel-thomas, line 94: Seeding in tensorflow is a bit more complex, I've given up trying to get fairly reproducible results for tests -- the non-determinism of some ops doesn't help ^^

https://www.tensorflow.org/api_docs/python/tf/random/set_seed
https://github.com/tensorflow/tensorflow/issues/2732

tests/utils.py
```diff
@@ -50,3 +52,57 @@ def generate_regression_model(features_shape, output_shape=1):
...
+
+        # create one dense layer
+        tf_model = Sequential()
+        tf_model.add(tf.keras.layers.Dense(1))
+
+        # compile model
+        tf_model.compile(optimizer="adam", loss="mae")
+
+        # train model
+        tf_model.fit(X_train, y_train, epochs=50, validation_split=0.25, verbose=0)
```
fel-thomas, line 105: I would clearly avoid this in the test pipeline, we don't want to train models, especially linear model.
For the metric test, influence function and Sobol test I had the same problem and decided to put fixed weight / fixed equation and do the math to ensure we are correct (see deconvnet / guided backprop tests) ;)

fel-thomas (commented):
Super clean / clear, awesome job ! 
If possible, I suggest we avoid `.fit(.)` in the test pipeline -- I propose to set fixed weights in both models it would be quicker to run and the test would be more deterministic / accurate. What do you think @AntoninPoche ? ;)
Once it's done, I let you merge !

tests/generic/test_wrapper.py
AntoninPoche: Now `almost_equal()` is used

tests/utils.py
AntoninPoche: Good remark, I removed those training, now the weights are set manually to the dataset coefficients.

AntoninPoche:
Good remark @fel-thomas, I modified the code so that no fitting is used.

fel-thomas (approved):
LGTM !

PR #100, opened 2022-03-09
Correct documentation links in the Read.me

AntoninPoche:
# Correction of the home page documentation links

When I tried to access the documentation of attribution methods through the read.me page, the page did not load correctly. I tried with other documentation but it did not work.

Only the general link work, thus I recuperated the link though it and updated the one in the read.me (see 6aa85ba). I do not know if it the good way to do it. But in my view, it solves the problem.


fel-thomas (approved):
Well done ! It's important to fix that quickly, I think we should make a quick release to master now no ?

PR #107, opened 2022-08-11
Resolved `len is not well-defined for symbolic tensors`

TeodorChiaburu:
Replaced `len()` with `shape[0]` for several feature attribution methods (*lime*, *occlusion*, *rise* and *smoothgrad*) to fix error `len is not well-defined for symbolic tensors`


AntoninPoche (approved):
LGTM

fel-thomas:
LGTM

PR #110, opened 2022-08-25
Plots: introduce single attribution plot

fel-thomas:
# Fix
apparently, in commit there was still a text duplicate in the doc, 088f3b67f2afb32be2fc66a1b69e739701445938 fix it.

# Plot
Current plot function only allow to plot multiple explanations. To allow more flexibility I proposed to split the current function into multiple sub functions to introduce a sub_function that allow to plot only one explanation.

Also, a simple wording issue but the `standardization` function was in fact doing a normalization. I propose to rename it.

This function is going to be useful for the next notebook on 'Sanity check'. :)


AntoninPoche (approved):
LGTM, the separation of the two function is indeed useful !

PR #111, opened 2022-08-26
Preparing for HSIC Attribution Method

fel-thomas:
this PR introduces a refactoring in the Sobol module henceforth called Global Sensitivity Analysis so that code can be shared with the HSIC method soon to be introduced.

As a matter of fact, the estimators, sampler and the core of the method can be shared. 

## Fix

9c44321ff8aac46c7e454371774d5a1531fa163c reduce drastically the number of forward during sanity check tests (when we test the method can be called with correct args, should return a correct shape...) to speed up the tests.

ef79ee5115d7f2aa398f39ccbeb33822b5fe766f relax the strict equality constraint on an almost equal to take into account numerical issue when doing the tests on gpu (e.g on my new mac haha). 

## Preparing HSIC

We will soon introduce the new HSIC Attribution method. In order to avoid code redondance with Sobol as they are both Global Sensitivity Analysis methods, I propose an abstraction called `GSABaseAttributionMethod` that will be parametrized by an estimator and a sampler and take care of doing the forward and apply the perturbations.

But first, we need to rename the `sobol` module into `global_sensitivity_analysis` b797b3fa1c71c1dddb41331d6f4664cb96b76eaa.

 The `sampler` module now need to be more specific as for Sobol estimation we use a specific sampling mechanism
(replicated design), ac2677365032ae807ee3bccb199bad4a050ebcb9 rename and specify each `XXXsampler` into `XXXsamplerRS` for replicated design.

Similarly, the `estimator` module is now the `sobol_estimator`module 37b59f88f93bf2624573d967e682339ed266d9e2.

Now the big part, I abstracted the forward / perturbation mechanism of the `SobolAttributionMethod` and created a new base explainer for GSA based attribution 35e4f3a3015b9e86984975c9bcd2efda47c07508.

See how SobolAttributionMethod is super slim ! 
Also, good news It would be the same with HSIC ! ;)


AntoninPoche (changes requested):
Test optimizations are welcomed and the sobol rename LGTM.

But it seems to me that you are requesting a merge directly on master without putting you changes trhough develop, is it normal? You may have done that with prior PR.

lucashervier (approved):
LGTM too! But @AntoninPoche is right you are currently going to merge into master directly :)

paulnovello (approved):
Ok for me !

fel-thomas:
Thanks for the reviews, ok for `develop`. :)

PR #114, opened 2022-10-16
Add plot_attribution to __init__

TeodorChiaburu:
I was not able to import `plot_attribution` (the one without the 's' at the end, used for plotting single images). I just added the function's name to the `__init__` in **xplique.plots**. Now it can be imported.


fel-thomas:
LGTM, thank you very much Teodor !! :D

fel-thomas (approved):

PR #115, opened 2022-10-17
NeurIPS release: doc enhancement, new tutorial, rollback SmoothGrad

fel-thomas:
# Improving to build better later

This PR (and the next one) are dedicated to the preparation of NeurIPS 2022 where we will try to present Xplique again (using our 3 posters). Therefore more about cleaning than adding functionality.

7edae63fd4c6a985610703d9c496bc9b56ba4a1b fix duplicate in the citation section. 
8126c2e75c81e11b82706579e7bde6349885cf5e updating MathJax in the documentation (faster loading yeah!)
95683347ead81b70a9f0bb670b3ed59b385e21ab bunch of design improvement
475d57d6417f1063cb4a3e381ceedef0045a759c removing all auth in the colabs links
b1d14bb76906c83346eab8de4cb5a9d8889e6c6f major revision of bunch of pages, following the new style proposed by 
@AntoninPoche and @lucashervier (btw I added your picture there bf3c510a62a1de8c9bf62201ea974fbceaf8ae41) ^^

# SmoothGrad case

8eee7090abf4da149e07ffa89c5c6ef86e21a1b8 is an important commit. Please, if you could do a double check, that would be great. Smoothgrad needs to store all points (N samples) to apply the reduction (either mean, variance or mean squared). To do this, we proposed to do it online in a previous commit. As a result, VarGrad and SquareGrad were obsolete (and there were prints in the code). I propose a more classical algorithm where we need at least the memory for the tensor (N, *input_shape). I think that if a user can't store this tensor, then Xplique shouldn't have to deal with this kind of memory optimization. The code for inline aggregation would be complex and add little value.

In the future, we could look at https://github.com/online-ml/river library, but that would require some time. :)


xplique/attributions/smoothgrad.py
```diff
@@ -73,45 +73,72 @@ def explain(self,
...
-                    [smoothed_gradients, reduced_reduced_gradients[tf.newaxis, :]], axis=0)
+        Returns
+        -------
+        noisy_inputs
+            Duplicated inputs with noisy mask applied.
+        """
+        nb_samples = len(noisy_mask)
 
-        return smoothed_gradients
+        noisy_inputs = tf.repeat(tf.expand_dims(inputs, axis=1), repeats=nb_samples, axis=1)
```
AntoninPoche, line 137: Do we agree that this creates a Tensor of shape `(batch_size, nb_samples, *inputs.shape[1:])`? For high resolution images, the `nb_sample` parameter should be higher than usual image, combined with the important `inputs.shape`, does it not create a really huge tensor even with a `batch_size` of 1?

My point here is that, for me there may be models or dataset for which the tensor created with `nb_samples` necessary for convergence cannot be stored in the memory.

AntoninPoche (commented):
Great update of the documentation, I am just waiting for some answers on SmoothGrad. Otherwise, LGTM, thank you Thomas, as alwais, great work !

AntoninPoche (approved):

Issue #120, opened 2022-12-05
[Enhancement] - Your suggestion/feature request


lucashervier:
**Is your feature request related to a problem? Please describe.**
Workflow dependant on Python 3.6 is deprecated. We should however cover Python 3.9 and maybe 3.10

**Describe the solution you'd like**
Update workflow

fel-thomas:
What about adding it to the matrix in the CI ? :)

AntoninPoche:
Treated in #132.

PR #119, opened 2022-11-27
Hsic Attribution Method

fel-thomas:
## Summary

This PR introduce the work from _Novello, Fel, Vigouroux_ 
**Making Sense of Dependence**: Efficient Black-box Explanations Using Dependence Measure https://arxiv.org/abs/2206.06219 (NeurIPS 2022).

This new black-box attribution method will be part of the GSA module.

To introduce HSIC, we need 3 key ingredients: samplers, kernels and HSIC Estimator. We do that in 3 separate commits and then introduce the Attribution method by exposing the `HSICAttributionMethod` API.

## HSIC Attribution method

42e2602defe60d625236c0780805b0d56bc1a35e introduce the samplers, c56e9b3074236b8c88e8af7f6deeeeb08fabdad3 the kernels  and 8eeab031ec10ffa8bbb2f1376756d425ca991ccd use them to estimate HSIC.

c9a9e0f4f79df62e3c31b16a7fe4f0dc2b6038d3 modify the abstract `GSAAttributionMethod` class to naturally derive HSIC in the next commit 4be2fa2aef1b41432df1852892ea12b5ad89ab42.

PR #121, opened 2022-12-05
Change Python distribution

lucashervier:
## Update Python version for CI
Commit 217fb77 adress #120.

## Docs
 27d48d3 allow to justify content in the generated documentation

## Templates
4412551 modify issues template for a more user-friendly one


AntoninPoche (approved):

fel-thomas:
We discussed that issue with Lucas yesterday and in fact I already adress it in the previous PR for HSIC #119. 

Many people are on python `3.6` and I sincerely think it's too early to remove it. Actually the problem is the CI Pipeline which is running on `ubuntu-latest`,  which makes the tests unstable. 

I propose to fix the version of Ubuntu at 20.04 which is still used a lot and to increase this version every year.

If you validate I'll close it. :)

PR #124, opened 2023-02-02
Fel/operator: Operator PR

fel-thomas:
# Operator(s) Abstraction

This PR introduces the concept of an operator, which aims to generalize attribution methods for a wide range of use-cases. The idea is as follows: to define an attribution method, we need any function that takes in the model (`f`), a series of inputs (`x`) and labels (`y`) and returns a scalar in `R`.

**`g(f, x, y) -> R`**

This function, called an operator, can be defined by the user (or by us) and then provides a common interface for all attribution methods that will call it (or calculate its gradient). As you can see, the goal is for attribution methods to have this function as an attribute (in more detail, this will give `self.inference_function = operator` at some point).

# Some Examples of Operators

1) The most trivial operator is perhaps the **classification** one, it consists of taking a particular logit to explain a class. In the case where the model `f: R^n -> R^c` with `c` being the number of classes and `y` being one-hot vectors, then our operator simply boils down to:

```python
def g(f, x, y):
    return tf.reduce_sum(f(x) * y, -1)
```

2) Regarding **segmentation**, with a model `f: R^w×h×c -> R^w×h×r` with `r` being the number of channels, there is no properly defined operator in the literature for segmentation, but we could imagine explaining a whole channel, with `y` being a one-hot vector that explains channel `r`, the operator therefore simply boils down to:

```python
def g(f, x, y):
    return tf.reduce_sum(f(x) * y[:,None,None,:], (1,2))
```

3) Regarding **bounding-box**, an operator has already been defined in the literature with the D-RISE article. It consists of using the three IOU, objectness, and box classification scores to form... *a scalar*!

4) To explain **concepts**, for example with a model `f = c ⚬ g(x)`, with `a = g(x)` and a factorizer  that allows interpreting `a` in a reduced dimension space `u = factorizer(a)`, we can very well define the following operator:

```python
def g(c, u, y):
    a = factorizer.inverse(u)
    y_pred = c(a)
    return tf.reduce_sum(y_pred * y, -1)
```

As you can see, many cases can be handled in this manner!

# Implementation

Regarding implementation, there is a series of operators available in the file in `commons/operators` and the most important part -- the operator plug -- is located in the `attributions/base.py` file. As discussed with @AntoninPoche  & @lucashervier, I think you know where I'm coming from but the PyTorch implementation is not far and would be located here!

Once this is done, I simply added the argument to all the attribution methods defined in the library.

This being a quite important PR (not in terms of code line but in terms of internal API change), I'm not against a careful reading and re-reading by the team! :)


xplique/metrics/fidelity.py
```diff
@@ -9,7 +9,7 @@
 from scipy.stats import spearmanr
 
 from .base import ExplanationMetric
-from ..commons import batch_predictions_one_hot
+from ..commons import batch_predictions
```
AntoninPoche, line 12: In the whole fidelity file, `batch_predictions` is used, however, it corresponds to the default classification operator. In this sense, metric would be limited to classification.

I suggest that metric also take `operator` as an `__init__` argument and share the same code as BlackBoxExplainer in their init to define the `self.batch_predictions`. A quick patch could be to instanciate a dummy explainer and use its arguments.

Nonetheless, I think that in the long term, that all insertion metrics are merge into one supporting several data types and task. (Same for deletion).

AntoninPoche (changes requested):
This is great work! I think it is more elegant and flexible than previously. I also like the new exception raised for gradient-based methods.

However, I think we should update the metrics.

AntoninPoche:
I am concerned about how this PR would support multi-output regression.

For me what we want for perturbation based-methods is to compare the initial pred to the modified pred on all outputs. In this sense, we cannot reduce the output of the inference function to a scalar. Nonetheless, methods like Occlusion, when they aggregate sensitivity, they do not take this increase of dimension into account. I have no quick fix for this. And I do not think that taking the mean of the output is pertinent.

However, gradient-based functions on multi-output regression task want the output to be reduced, otherwise, they would output one gradient for each output.

Finally, we could reduce the output of the inference function for all methods if we take into account the ground truth. We can say that the inference function output is the mae for example.

What do you think?

Agustin-Picard (commented):
It might be worth it to define an interface for these operators, or at least, specify more clearly the types of the callable (ex. `Callable[[tf.keras.Model, tf.Tensor, tf.Tensor], tf.Tensor]`). This should allow us to perform the check that Antonin's asking for.

xplique/attributions/base.py
```diff
@@ -44,28 +46,51 @@ class BlackBoxExplainer(ABC):
...
-            self.model = model
-            self.inference_function = predictions_one_hot
-            self.batch_inference_function = batch_predictions_one_hot
         else:
             self.model = model
+
+        if operator is not None:
+            # user specified a custom operator, we wrap it to generate the batching operator
+            # and the gradient function of this operator
+            self.inference_function = operator
```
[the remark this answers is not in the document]
lucashervier: I would add that the description is really vague in the docstring.

xplique/attributions/base.py
```diff
@@ -44,28 +46,51 @@ class BlackBoxExplainer(ABC):
...
+            # maybe a custom model (e.g. tf-lite), we can't backprop through it
+            self.inference_function = predictions_operator
+            self.batch_inference_function = batch_predictions
+            self.gradient = self.batch_gradient = no_gradients_available
+
+        else:
+            # completely unknown model (e.g. sklearn), we can't backprop through it
             self.inference_function = predictions_one_hot_callable
             self.batch_inference_function = batch_predictions_one_hot_callable
+            self.gradient = self.batch_gradient = no_gradients_available
```
lucashervier, line 93: I think it is not a really clean way to do it (double "=" in the same line.

xplique/attributions/base.py
```diff
@@ -44,28 +46,51 @@ class BlackBoxExplainer(ABC):
...
+            self.inference_function = predictions_operator
+            self.batch_inference_function = batch_predictions
+            self.gradient = gradients_predictions
+            self.batch_gradient = batch_gradients_predictions
+
+        elif isinstance(model, (tf.Module, tf.keras.layers.Layer)):
+            # maybe a custom model (e.g. tf-lite), we can't backprop through it
+            self.inference_function = predictions_operator
+            self.batch_inference_function = batch_predictions
+            self.gradient = self.batch_gradient = no_gradients_available
```
lucashervier, line 87: I think it is not a really clean way to do it (double "=" in the same line.

xplique/commons/operators.py
```diff
@@ -0,0 +1,140 @@
...
+    operator
+        Any callable that take model, inputs and labels as parameters.
+
+    Returns
+    -------
+    batched_operator
+        Function that apply operator by batch.
+    """
+
+    def batched_operator(model, inputs, targets, batch_size=None):
```
lucashervier, line 123: I think we could do something more efficient with `map` or `map_fn` rather than using a for loop

lucashervier (commented):
Some minor remarks concerning optimization and documentation!

AntoninPoche:
Hello guys, I will summarize our discussion on this PR and list things to change:

- [ ] Implement a `check_operator` function used in `BlackBoxExplainer.__init__()`. This function should check that the operator follows the expected format and guide user through exceptions overwise.
- [ ] The operator for `BlackBoxExplainer` do not need to be differentiable, hence the operator gradient management can be transferred to `WhiteBoxExplainer`.
- [ ] The documentation of operator should be clearer in the docstring of functions.
- [ ] The bug on feature viz should be corrected as we do not want to remove possibilities from the library.
- [ ] Metrics should use the operator, we will add it as an argument of the metrics. We should be careful to check if the operator is in fact an operator through the aforementioned check. (I will take care of that one and merge the two `CausalFidelity` metrics).

The following points were deemed as further enhancement that will be taken into account in future PR:
- Go from [3.6, 3.8] to [3.7, 3.10]
- Update Readme
- Cleaner adaptation of BoundingBox to the operator API.
- Support Pytorch.
- Tutorial on operator and target at the same time.
- Other tutorials
- NLP
- Diffusion models

xplique/attributions/base.py
```diff
@@ -3,15 +3,42 @@
...
+        True if the operator is valid, False otherwise.
+    """
+    is_callable = isinstance(operator, Callable)
+    args = inspect.getfullargspec(operator).args
+
+    # we allow operator with optional arguments, but the first 3 must be present
+    is_valid = is_callable and len(args) >= 3
+
+    if not is_valid:
+        raise_invalid_operator()
```
AntoninPoche, line 41: This function should not be in `attributions/base.py`, but in the operator file, i.e. `commons/operator.py`.

AntoninPoche (commented):
Just one last modififcation, `check_operator` function should be moved in my view. Otherwise LGTM

AntoninPoche (approved):

xplique/commons/operators.py
fel-thomas: I agree that `map_fn` and `vectorized_map` would help for parallelization, but why do you want that for inference (just increase batch_size)? ^^

Still, I think that could be an interesting tool if we had preprocessing / data augmentation!

xplique/attributions/base.py
fel-thomas: You're right (same suggestion @Agustin-Picard), I have added a better typing.
Also, we discussed with @lucashervier and @AntoninPoche but we plan to dedicate a page on the documentation (replace the callable page) to dive into this `operator` API.
Last but not least, I want to create few notebooks to showcase the possibilities of this framework.
If you have other ideas, don't hesitate ;)

PR #125, opened 2023-02-23
Adapt metrics to the Operator PR

AntoninPoche:
# Adapt metrics to the Operator PR [#124]

The operator update is huge, hence we decided to devide it. This pull request concerns the adaptation of metrics to the operator update. (I do not know if this pull request should be merge into `fel/operator` branch or `master`).

The modifications can be divided in two parts some operator modifications and the adaptation of metrics to the operator PR:

# Update on the operator

- ### Move the model switch-case [commit 7f703c7]

In PR [#124] in `xplique/attributions/base.py`, in the `__init__` method of `BlackBoxExplainer` there was a switch case to define the methods `self.inference_function` and `self.batch_inference_function`. However, I had to do the same switch case in the metric base file. Hence I move it to `xplique\commons\operators.py` in the form of a function.

I did the same for `self.gradient` and `self.batch_gradient` in `WhiteboxExplainer` to be consistent. Furthermore, I added one case to be sure to detect non-backpropagable models and assign them `no_gradients_available`.

- ### Correct the tf.function bug [commit be834e2]

Through the metrics test, I discovered that when `check_operator` was called on an operator decorated by `@tf.function`, the line to get the function's arguments crashed. With the suggestion of @fel-thomas, we tested if the operator was a tf function and extracted its pythonic version for the test.

I also added an operator for regression models when someone wants to know the attribution on all outputs simultaneously. `regression_operator`.

- ### Update operator tests [commit d7c0493]

The previous commit highlighted that further tests were needed for operators. Hence I tested that our proposed operators are indeed operators and that wrong values for operators are detected.

# Update fidelity metrics 

- ### Add operator API to fidelity metrics [commit b42531d]

The operators were added to the `ExplanationMetric` class from which inherit fidelity metrics. `AverageStability` use the `explainer` which already includes the new operators and was not modified.  `MeGe` needed much more modifications and was not inheriting from base metric API, therefore I did not update it.

The metrics inheriting from `ExplanationMetric` have naturally been updated in their constructor and now use the operator in their `evaluate()` method. (Namely `MuFidelity`, `Insertion`, `Deletion`)

- ### Update common and fidelity tests [commit 679dc67]

We now make sure that the fidelity metrics support an operator such as `regression_operator` for a regression model.

- ### Update the documentation [commit 394ad77]

The documentation of the API did not include the operator description and the interpretation of the metrics was too specific to classification.

- ### Remove redundant metrics [commit 5ca8aeb]

Previously, metrics `InsertionTS`, `InsertionTab`, `DeletionTS`, and `DeletionTab` were created because the metrics were too specific to classification and did not support regression. We could say that the now default `operator` was hardcoded in the metrics.

These classes were naturally removed in this pull request. We could argue to let deprecated warnings, but I do not think anyone used them.


docs/api/metrics/deletion.md
```diff
@@ -1,7 +1,6 @@
 # Deletion
 
-The Deletion Fidelity metric measures how well a saliency-map–based explanation of an image classification result localizes
-the important pixels.
+The Deletion Fidelity metric measures how well a saliency-map–based explanation localizes the important features.
```
lucashervier, line 3: This description would be true for Insertion too. I think we could specify "by iteratively deleting the most important features while tracking the prediction score" or something like that!

lucashervier: Okay considering what follows maybe just saying "a score" would do

tests/metrics/test_fidelity.py
```diff
@@ -44,64 +45,48 @@ def test_causal_metrics():
...
+    }
+
+    for type, input_shape in data_types.items():
+        nb_labels, samples = 10, 20
+        x, y = generate_data(input_shape, nb_labels, samples)
+        model = generate_regression_model(input_shape, nb_labels)
+
+        explanations = np.random.uniform(0, 1, x.shape[:-1])
+    
+    # test several
```
lucashervier, line 64: I do not know if its a mistake or not but here you say you will test several operators but only one is in the dictionnary.

AntoninPoche: Initially, there was also the `prediction_operator` however, for regression model, it could output negative values. Hence, fidelity metrics could also output negative values. Which does not correspond to the expected behavior of the metrics.

Hence it is a mistake on my part. Nonetheless, it leads me to the question, should we take the absolute value of the operator output in the metrics.

lucashervier (changes requested):
Looks good to me. I would maybe give more details on Deletion documentation but that is not really a huge need. However, there is a test where we expect several operators and only one is challenged. Is it a comment error or some operators are really missing ?

tests/metrics/test_fidelity.py
```diff
@@ -1,8 +1,10 @@
 import tensorflow as tf
 import numpy as np
 
-from ..utils import generate_model, generate_timeseries_model, generate_regression_model, generate_data, almost_equal
-from xplique.metrics import Insertion, Deletion, MuFidelity, InsertionTS, DeletionTS, InsertionTab, DeletionTab
+from ..utils import (generate_model, generate_regression_model, generate_timeseries_model, 
+                     generate_data, almost_equal)
+from xplique.commons.operators import predictions_operator, regression_operator
```
lucashervier, line 6: Unnecessary `predictions_operator`

lucashervier (approved):

lucashervier (approved):
LGTM

PR #126, opened 2023-04-20
fix: HSIC hotfix

paulnovello:
* doc of __call__ of HSICEstimator
* Add @tf.function


AntoninPoche (approved):

fel-thomas:
Please rebase to have only 1 commit, and rename it to something like :
"hsic attribution: refactor estimator for tf.function" ? (check the other commits name)   ;)

PR #133, opened 2023-07-20
Attributions: introduce FORGrad 

fel-thomas:
### Gradient strikes back: How filtering out high frequencies improves explanations

This PR contains the implementation of FORGrad (paper here: https://arxiv.org/pdf/2307.09591.pdf). All in all, this method consists in filtering the noise in the explanations to make them more interpretable.

Basically, it use low-pass filter on the heatmaps after performing any of the attribution methods.
It therefore doesn’t require any major challenge, just a couple of functions will make the explanations look better.

### Core

Commit a7a0cb17bd7776590d121c06a3c98bb7a630ede9 contains the code with the implementation and the test. FORGrad is in the Attribution method but it is an attribution method enhancement. Otherwise, it has to be used in complement to a “standard” attribution method.

### Docs and utils

Commit 08f77478f47a4cabe792032983fa47f33b218fef provides the documentation and attach the notebook to the readme and documentation.

we (@S4b1n3, @fel-thomas) request the review from @lucashervier and @AntoninPoche.

Here is the notebook :) 
https://colab.research.google.com/drive/1ibLzn7r9QQIEmZxApObowzx8n9ukinYB#scrollTo=JJWlCx6LdXOH

Main author: @S4b1n3 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ We propose some Hands-on tutorials to get familiar with the library and its api:
 
 - [**Attribution Methods**: Sanity checks paper](https://colab.research.google.com/drive/1uJOmAg6RjlOIJj6SWN9sYRamBdHAuyaS) <sub> [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1uJOmAg6RjlOIJj6SWN9sYRamBdHAuyaS) </sub>
 - [**Attribution Methods**: Tabular data and Regression](https://colab.research.google.com/drive/1pjDJmAa9oeSquYtbYh6tksU6eTmObIcq) <sub> [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1pjDJmAa9oeSquYtbYh6tksU6eTmObIcq) </sub>
+- [**FORGRad**: Gradient strikes back with FORGrad](https://colab.research.google.com/drive/1ibLzn7r9QQIEmZxApObowzx8n9ukinYB) <sub> [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1ibLzn7r9QQIEmZxApObowzx8n9ukinYB) </sub>
 - [**Attribution Methods**: Metrics](https://colab.research.google.com/drive/1WEpVpFSq-oL1Ejugr8Ojb3tcbqXIOPBg) <sub> [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1WEpVpFSq-oL1Ejugr8Ojb3tcbqXIOPBg) </sub>
 
 <p align="center" width="100%">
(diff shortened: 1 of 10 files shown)


xplique/attributions/forgrad.py
```diff
@@ -0,0 +1,90 @@
...
+
+import tensorflow as tf
+import numpy as np
+
+
+@lru_cache(maxsize=32)
+def _low_pass_real_signal_mask(size : int, bandwith : int) -> tf.Tensor:
+    """
+    Create a low pass filter mask to be applied on a real signal.
+    Since the DFT of a real signal is Hermitian-symmetric, only returns the fft_length / 2 + 1
```
lucashervier, line 15: Can we use the real acronym meaning, at least the first time ? (DFT, FFT)

xplique/attributions/forgrad.py
```diff
@@ -0,0 +1,90 @@
...
+    Parameters
+    ----------
+    size
+        Size of the mask.
+    bandwith
+        Bandwith of the low pass filter. The higher the bandwith, the more frequencies are kept.
+
+    Returns
+    -------
+    mask
```
lucashervier, line 27: Am I right if I say that this object has always the same number of dimension ? In that case, it could be nice to specified the output's shape.

xplique/attributions/forgrad.py
```diff
@@ -0,0 +1,90 @@
...
+    # un-center the mask
+    mask = tf.signal.fftshift(mask)
+    # keep only the unique components
+    mask = tf.cast(mask, tf.float32)[:, :size//2 + 1]
+
+    return mask
+
+def forgrad(explanations : tf.Tensor, sigma: int = 15) -> tf.Tensor:
+    """
+    ForGRAD is a method that enhances any attributions explanations (particularly useful on
```
lucashervier, line 45: As it works on "any attributions explanations" I would have put it in the `commons` folder as we did for the `operators`. I guess it is debatable but in my opinion, it would make sense to be in the `attributions` folder you should inherit from `base` (or another attribution method). But here, it is really a standalone function that we can add "on-top".

README.md
```diff
@@ -81,6 +81,7 @@ We propose some Hands-on tutorials to get familiar with the library and its api:
 
 - [**Attribution Methods**: Sanity checks paper](https://colab.research.google.com/drive/1uJOmAg6RjlOIJj6SWN9sYRamBdHAuyaS) <sub> [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1uJOmAg6RjlOIJj6SWN9sYRamBdHAuyaS) </sub>
 - [**Attribution Methods**: Tabular data and Regression](https://colab.research.google.com/drive/1pjDJmAa9oeSquYtbYh6tksU6eTmObIcq) <sub> [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1pjDJmAa9oeSquYtbYh6tksU6eTmObIcq) </sub>
+- [**FORGRad**: Gradient strikes back with FORGrad](https://colab.research.google.com/drive/1ibLzn7r9QQIEmZxApObowzx8n9ukinYB) <sub> [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1ibLzn7r9QQIEmZxApObowzx8n9ukinYB) </sub>
```
lucashervier, line 84: Is it really belonging here ? All other tutorials concerning a specific approach (Rise, Saliency, etc...) are only referenced in the table below while here it is more "Getting Started" stuff.

lucashervier (changes requested):
This is really neat. Nice work! 🔥  My only true question is more concerning where to put this `forgrad.py`. I shared my opinion in the different comments and I would like to know what you think about it

S4b1n3:
After discussing with Thomas, we agree with all the comments.
We will update the commits: specifying DFT and the shapes, as well as changing the location of the method. 
Thanks for the feedback!

fel-thomas:
Forgrad is now in common, @lucashervier is right it make more sense.
Also, the doc is expanded as asked. 

I think we are ready to merge ;) 


Big thank @lucashervier <3

lucashervier (approved):
LGTM ;) Well done you two!

PR #149, opened 2023-11-27
Cockatiel

fredericboisnard:
First PR about the integration of Cockatiel in Xplique.

### DIFF
--- a/tests/concepts/test_cockatiel.py
+++ b/tests/concepts/test_cockatiel.py
@@ -1,0 +1,240 @@
+import numpy as np
+import torch
+import torch.nn as nn
+
+from torch.nn import MSELoss
+from transformers import RobertaPreTrainedModel, RobertaModel
+from transformers import RobertaTokenizerFast
+from xplique.concepts import CockatielTorch as Cockatiel
+from xplique.commons.torch_operations import NlpPreprocessor
+
+
+class FakeImdbClassifier(torch.nn.Module):
+    def __init__(self, nb_classes):
+        super().__init__()
+        self.nb_classes = nb_classes
+
+    def features(self, **kwargs):
+        start_value = 0
+        return torch.arange(start_value,
+                            start_value + kwargs['input_ids'].shape[0] * self.nb_classes).\
+            reshape(kwargs['input_ids'].shape[0], self.nb_classes)
+
+    def classifier(self, latent):
+        # Simulates a classifier, returns alternatively [0, 1] and [1, 0]
+        # as result
+        n = len(latent)
+        pattern = torch.cat([torch.tensor([0, 1]), torch.tensor([1, 0])])
+        ypred = torch.cat([pattern] * n).reshape(n * 2, -1)
+        return ypred
+
+    def forward(self, **kwargs):
+        return self.classifier(self.features(**kwargs))
+
+
+class ImdbPreprocessor(NlpPreprocessor):
+    def preprocess(self, inputs: np.ndarray, labels: np.ndarray):
+        preprocessed_inputs = self.tokenize(samples=inputs.tolist())
+        preprocessed_labels = torch.Tensor(np.array(
+            labels.tolist()) == 'positive').int().to(self.device)
+        return preprocessed_inputs, preprocessed_labels
+
+
+def test_shape():
+    """Ensure the output shape is correct"""
+
+    device = torch.device(
+        "cuda") if torch.cuda.is_available() else torch.device("cpu")
+    pretrained_model_path = "wrmurray/roberta-base-finetuned-imdb"
+    tokenizer = RobertaTokenizerFast.from_pretrained(pretrained_model_path)
+    nb_classes = 2
+    model = FakeImdbClassifier(nb_classes=nb_classes)
+    model = model.eval()
+
+    imdb_preprocessor = ImdbPreprocessor(tokenizer,
+                                         device,
+                                         padding="max_length",
+                                         max_length=512,
+                                         truncation=True,
+                                         return_tensors='pt')
+
+    input_to_latent_model = model.features
+    latent_to_logit_model = model.classifier
+    number_of_concepts = 20
+    cockatiel_explainer_pos = Cockatiel(input_to_latent_model=input_to_latent_model,
+                                        latent_to_logit_model=latent_to_logit_model,
+                                        preprocessor=imdb_preprocessor,
+                                        number_of_concepts=number_of_concepts,
+                                        batch_size=64,
+                                        device=device)
+
+    data = ["Absolutely riveting from beginning to end. Test 1 2 3. Another test.",
+            "But it's a great movie.",
+            "This is the best movie of the year to date.",
+            "The movie is excellent."]
+    crops, crops_u, concept_bank_w \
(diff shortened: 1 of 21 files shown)


fredericboisnard:
The conflicts must come from the 1st patch "test moving the EPSILON & similar to free some RAM" -> I will remove it from the PR since it was mostly to try to improve RAM management on my side (not sure if we want to take it ; and if yes, then it would be better in another PR I suppose)

xplique/commons/nlp.py
```diff
@@ -0,0 +1,288 @@
...
+
+        Returns
+        -------
+        clause_list
+            A list with input texts split into clauses.
+        separator
+            The separator used for spliting the sentence.
+        """
+        if not isinstance(sentence, str):
+            sentence = '.'.join(sentence)
```
AntoninPoche, line 184: Same remark.

xplique/commons/nlp.py
```diff
@@ -0,0 +1,288 @@
...
+            return FlairClauseExtractor(tagger, clause_type)
+        if extract_fct == "spacy_clause":
+            return SpacyClauseExtractor(pipeline, clause_type)
+        if extract_fct == "sentence":
+            return SentenceExtractor()
+        if extract_fct == "word":
+            return WordExtractor()
+        if extract_fct == "excerpt":
+            return ExcerptExtractor()
+        raise ValueError("Extraction function can be only 'clause', \
```
AntoninPoche, line 287: You did not include `"flaire_clause"` and `"spacy_clause"` in the error message, but you left `"clause"` which is not a possibility.

xplique/commons/nlp.py
```diff
@@ -0,0 +1,288 @@
...
+
+        return clause_list, self.separator
+
+
+class ExtractorFactory():
+    """
+    Factory for extractor classes.
+    """
+    @staticmethod
+    def get_extractor(extract_fct="sentence", tagger=None, pipeline=None, clause_type=None):
```
AntoninPoche, line 255: I think that `language` should be an argument of this function.

xplique/commons/operators.py
```diff
@@ -7,7 +7,7 @@
 import tensorflow as tf
 
 from ..types import Callable, Optional
-from ..utils_functions.object_detection import _box_iou, _format_objects, _EPSILON
```
AntoninPoche, line 10: I agree that the epsilon should not be loaded simultaneously as loading operators. But is it a good thing to initialize it at each run of the tf function?

xplique/commons/torch_operations.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class NlpPreprocessor(ABC):
+    """
+    Abstract base class for NLP preprocessing.
+
+    Parameters
+    ----------
+    tokenizer
+        The tokenizer function to be used for tokenization. It must be
```
AntoninPoche, line 20: This description needs to be clarified as no model is defined here. Does this correspond to what you define in the nlp_commons?

If it corresponds to the tokenizer-model pair, as in HuggingFace, then I recommend having a class that treats them both at the same time. It is easier to understand and the API would be more natural.

xplique/commons/torch_operations.py
```diff
@@ -0,0 +1,150 @@
...
+
+from abc import ABC, abstractmethod
+from math import ceil
+import numpy as np
+import torch
+
+from ..types import Union, Tuple, Callable, List, Dict
+
+
+class NlpPreprocessor(ABC):
```
AntoninPoche, line 13: This class is called NLP... Should it not be in the nlp commons?

AntoninPoche (changes requested):
I only read part of the PR for now. We will have to discuss the architecture of the code. It seem that cockatiel forces us to introduce nlp attributions.

tests/concepts/test_cockatiel.py
```diff
@@ -0,0 +1,240 @@
...
+    assert concept_bank_w.shape[0] == number_of_concepts
+
+    global_importance_pos = cockatiel_explainer_pos.estimate_importance()
+    assert len(global_importance_pos) == number_of_concepts
+    most_important_concepts_ids = global_importance_pos.argsort()
+
+    nb_excerpts = 2
+    nb_most_important_concepts = 2
+
+    best_sentences_per_concept = cockatiel_explainer_pos.get_best_excerpts_per_concept(
```
AntoninPoche, line 235: How can you be sure of which sentence will correspond to which concept?

AntoninPoche (changes requested):
Great work Fred, this is a huge PR! A complex one at that, which raises several points where we should all agree before merging. I listed several general remarks:
- No description of the PR in general, it can be great to clarify your choices and what you are trying to do.
- No notebook or tutorials, I cannot review the plot function or the API without examples. Furthermore, it will be necessary for the final version.
- I cannot launch the tests (from github), I think it is because of the conflicts.
- You did not base your code on the last version of master (1.3.3). I think that the problems you have with memory were solved with last pull requests. It should solve the conflicts at the same time.
- You did not make the documentation, it might be better to discuss the choice made before making it. It will save you some time.
- You remade a lot of indentations (automatically, I think), but it created many weird code parts. The line limit is 100 in Xplique.
- We will have to discuss two choices:
    - Where do we put the nlp parts? In my view a xplique.npl module is pertinent
    - The structure of the code (multi-inheritance)
- The code you added introduced a lot of dependencies to other libraries (flair, nltk, transformers (for tests)); we should discuss how to treat this. Maybe have an additional nlp_requirements.txt file to install when calling xplique.nlp? Furthermore, we should check the licenses of these libraries, they should be MIT, otherwise we cannot use them.

xplique/concepts/cockatiel.py
```diff
@@ -0,0 +1,569 @@
...
+        The batch size for all the operations that use the model. Default is 256.
+    device
+        The type of device on which to place the torch tensors
+    """
+
+    def __init__(
+            self,
+            input_to_latent_model: Callable,
+            latent_to_logit_model: Callable,
+            preprocessor: NlpPreprocessor,
```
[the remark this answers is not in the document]
fredericboisnard: Yes I understand your concern ; it's possible, but one of the goals of the NlpPreprocessor is to gather both the tokenizer and its parameters in one place, in order to 'proxy' the calls to the tokenizer. Then we can use this feature in cockatiel but also in external parts such as nlp_batch_predict():
- cockatiel._latent_predict() calls self.preprocessor.tokenize(), which will pass the good arguments to the tokenizer
- in torch_operations.nlp_batch_predict(preprocessor): we use preprocessor.preprocess() which again calls preprocessor.tokenize() to pass the good arguments to the tokenizer

If we pass the tokenizer alone in cockatiel.init(), then we will have to provide a variable number of parameters in addition (tokenizer arguments, can vary fron one tokenizer to another), and duplicate these parameters in nlp_batch_predict() :/

Or perhaps there is a better / easier way to implements this ?

Maybe I should rename the NlpPreprocessor into something like TokenizerHelper to better show it's simply to encapsulate tokenizer calls ?

Agustin-Picard:
Closing because stale.

COCKATIEL was implemented in the [Interpreto](https://github.com/FOR-sight-ai/interpreto) library.

Issue #151, opened 2023-12-08
[Bug]: - MuFidelity does not work for Tabular data and time series


AntoninPoche:
### Module

None

### Current Behavior

When applying MuFidelity on tabular data or time series, the number of features set to 0 with the mask does not correspond to `subset_percent`. Indeed, the formula to compute the `subset_size` depends on `grid_size ** 2` which is adapted to images.

Therefore, the number of features set to zero is much higher than expected, all features may be set to zero if `grid_size` is big. For tabular data or time series, the default value corresponds to the number of features, which is too big and results in fidelity scores of only zeros.

### Expected Behavior

Remove the attribute `self.subset_size` line 91.

Then in lines 201, 208, and 220:
Replace: `subset_masks = tf.argsort(subset_masks, axis=-1) > self.subset_size`
By: `subset_masks = subset_masks > self.subset_percent` 
!!! See the direction of the > or < depending on the meaning of subset percent.

### Version

1.3.1

### To Reproduce

See Attributions: Time Series and Regression tutorial, in the metric part, for `MuFidelity`, set `subset_percent` to `0.5` (perturb half) and `grid_size` to `48` (treat each time step differently).

It will return only zeros, but the configuration is far from absurd.

Issue #150, opened 2023-12-08
[Bug]: - Causal fidelity problem with `steps=-1`


AntoninPoche:
### Module

Metrics

### Current Behavior

When using a causal fidelity metric (either Insertion or Deletion) and setting `steps=-1`, the method returns an `Attribute error`.

### Expected Behavior

I expect the method to support parameter values indicated in the documentation.

To solve it, in fidelity.py:
Change L306: `tf.math.floor(self.nb_features * max_percentage_perturbed)`
By: `int(np.floor(self.nb_features * max_percentage_perturbed))`

### Version

1.3.1

### Relevant log output

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-8-ebb4e1c3761e> in <cell line: 7>()
      6 i = 0
      7 for explanation_name, explanations in explanations_to_test.items():
----> 8   deletion_score = metric(explanations)
      9   deletion_scores.append((explanation_name, deletion_score))
     10 

5 frames
/usr/local/lib/python3.10/dist-packages/xplique/metrics/base.py in __call__(self, explanations)
    157                  explanations: Union[tf.Tensor, np.array]) -> float:
    158         """Evaluate alias"""
--> 159         return self.evaluate(explanations)

/usr/local/lib/python3.10/dist-packages/xplique/metrics/fidelity.py in evaluate(self, explanations)
    326             better) curve.
    327         """
--> 328         scores_dict = self.detailed_evaluate(explanations)
    329 
    330         # compute auc using trapezoidal rule (the steps are equally distributed)

/usr/local/lib/python3.10/dist-packages/xplique/metrics/fidelity.py in detailed_evaluate(self, explanations)
    379         baselines_flatten = baselines.reshape(self.inputs_flatten.shape)
    380 
--> 381         steps = np.linspace(0, self.max_nb_perturbed, self.steps + 1, dtype=np.int32)
    382 
    383         if self.causal_mode == "deletion":

/usr/local/lib/python3.10/dist-packages/numpy/core/overrides.py in linspace(*args, **kwargs)

/usr/local/lib/python3.10/dist-packages/numpy/core/function_base.py in linspace(start, stop, num, endpoint, retstep, dtype, axis)
    118 
    119     """
--> 120     num = operator.index(num)
    121     if num < 0:
    122         raise ValueError("Number of samples, %s, must be non-negative." % num)

[...]

PR #152, opened 2023-12-12
Solve issues 150 and 151

AntoninPoche:
#150 was solved as suggested by casting to `int` before `np.linspace`.

#151 was solved as suggested by removing `self.subset_size` and doing computation with `self.subset_percent`.


Agustin-Picard (approved):
LGTM, awesome work, as usual @AntoninPoche 🔥

Issue #148, opened 2023-11-22
[Bug]: - Feature Visualization for single-channel image


tmt1611:
### Module

Feature Visualization

### Current Behavior

When using xplique.features_visualizations.optimize on model taking single-channel image as input, an dimension error occurs. 

I suspect this is caused by line 43 in xplique/features_visualizations/transformations.py : 

kernel = tf.tile(kernel, [1, 1, 3, 1])

where it should be kernel = tf.tile(kernel, [1, 1, number of channel, 1])

### Expected Behavior

xplique.features_visualizations.optimize return expected images visualizing features

### Version

1.3.0

### Environment

```markdown
- OS:linux
- Python version:3.11.3
- Tensorflow version:2.12.0
- Packages used version:
```

### Relevant log output

```shell
ValueError                                Traceback (most recent call last)
Cell In[8], line 12
     10 # create the objective, '-1' is the last layer, and the c_id's are the ids of the classes
     11 obj_logits = Objective.neuron(model, layer=-1, neurons_ids=[c_id for c_id, c_name in classes])
---> 12 imgs, _ = optimize(obj_logits,
     13                    nb_steps=1024, # number of iterations
     14                    optimizer=tf.keras.optimizers.Adam(0.05))
     16 # show the results
     17 plt.rcParams["figure.figsize"] = [12, 8]

File ~/.local/lib/python3.11/site-packages/xplique/features_visualizations/optim.py:112, in optimize(objective, optimizer, nb_steps, use_fft, fft_decay, std, regularizers, image_normalizer, values_range, transformations, warmup_steps, custom_shape, save_every)
    110 images_optimized = []
    111 for step_i in range(nb_steps):
--> 112     grads = optimisation_step(model, inputs)
    113     optimizer.apply_gradients([(-grads, inputs)])
    115     last_iteration = step_i == nb_steps - 1

File /usr/local/insa/anaconda/lib/python3.11/site-packages/tensorflow/python/util/traceback_utils.py:153, in filter_traceback.<locals>.error_handler(*args, **kwargs)
    151 except Exception as e:
    152   filtered_tb = _process_traceback_frames(e.__traceback__)
--> 153   raise e.with_traceback(filtered_tb) from None
    154 finally:
    155   del filtered_tb

File /tmp/__autograph_generated_filedrja5ayf.py:34, in outer_factory.<locals>.inner_factory.<locals>.tf__step(model, inputs)
     32     nonlocal imgs
     33     pass
[...]

PR #163, opened 2024-08-27
Feature Visualization: grayscale images

Agustin-Picard:
## Feature visualization for grayscale images

This PR adapts the code for generating feature visualizations (via Fourier, in pixel space, and using MaCo) for models trained on grayscale images.

In the case of grayscale images, no re-correlation is needed when optimizing in the Fourier domain (standard and MaCo). However, there is no standardized dataset that would generally work on any grayscale application (like ImageNet for natural images), so a dataset must be provided to compute the average magnitude for MaCo.

**Disclaimer:** Depending on the model under study, the optimization process can provide images that are really difficult to interpret. It seems to work much better on 1-Lipschitz models (trained using DEEL-LIP) and sometimes MaCo helps to generate better visualizations.

It responds to the feature requested in #148 

### Fix CI

The release of TF 2.16 integrates Keras 3.X, which seems to break some functionality. I've limited the installation of TF to <2.16.

Also, the release of NumPy 2.X breaks packages that were compiled using 1.X, and this includes TF. Thus, I've limited NumPy to 1.X.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
+numpy<2.0.0
 tensorflow>=2.1.0
-numpy
 scikit-learn
 scikit-image
 matplotlib
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ setup(
     author="Thomas FEL",
     author_email="thomas_fel@brown.edu",
     license="MIT",
-    install_requires=['tensorflow>=2.1.0,<2.16.0', 'numpy', 'scikit-learn', 'scikit-image',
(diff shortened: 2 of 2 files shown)


Agustin-Picard:
I had to limit tensorflow to versions <2.16 as the latest updates break a lot of things.

AntoninPoche (approved):
Looks good to me! Thank you very much Agus!

lucashervier (approved):
LGTM, I am merging :)

PR #165, opened 2024-09-25
Example-based

AntoninPoche:
# Introduce the Example-based module

This pull request introduces example-based methods from four families of methods. Namely:
- Similar Examples
- Semi-factuals
- Counterfactuals
- Prototypes

## The API

The API is common for the four families, and it has three steps:
- Construct a projection function with the model.
- Initialize the method with a dataset, projection, and other parameters.
- Explain given samples with a local explanation.

```
projection = ProjectionMethod(model)

explainer = ExampleMethod(
    cases_dataset=cases_dataset, 
    k=k,
    projection=projection,
    case_returns=case_returns,
    distance=distance,
)

examples = explainer.explain(inputs, targets)
```

PR #166, opened 2024-10-07
Example based

AntoninPoche:
This PR is the same as #165 but with a clean commit history.

# Introduce the Example-based module

This pull request introduces example-based methods from four families of methods. Namely:
- Similar Examples
- Semi-factuals
- Counterfactuals
- Prototypes

## The API

The API is common for the four families, and it has three steps:
- Construct a projection function with the model.
- Initialize the method with a dataset, projection, and other parameters.
- Explain given samples with a local explanation.

```
projection = ProjectionMethod(model)

explainer = ExampleMethod(
    cases_dataset=cases_dataset, 
    k=k,
    projection=projection,
    case_returns=case_returns,
    distance=distance,
)

examples = explainer.explain(inputs, targets)
```


lucashervier (approved):
Everything was already discussed and treated in PR #165 . As it is the same PR with cleaned commits I approve without further comments.

PR #176, opened 2026-01-25
Feature/fem

Zerg200:
Added the FEM method, added tests for the FEM method, and modified compatibility with the main codebase.


AntoninPoche (changes requested):
Great work @Zerg200 !

Most messages say that the FEM method is not at the right position in method listing. (alphabetic order).

The main thing to update is your method initialization.

Otherwise, everything LGTM

AntoninPoche:
Pull request redone in #180

Issue #177, opened 2026-01-29
[Bug]: tf.keras.Sequential error


quentinpossamai:
### Module

Attributions Methods

### Current Behavior

Using a tf.keras.Sequential (cf. image method 1) instead of a tf.keras.Model (method2) fails when caching the model in xplique.attributions.base.py L63 as .input will not be defined for Sequential objects (yes I have called the model once)

---

<img width="539" height="186" alt="Image" src="https://github.com/user-attachments/assets/0ee22550-8e40-40f6-82ef-7d34f513c3d9" />

---
<img width="589" height="103" alt="Image" src="https://github.com/user-attachments/assets/15e5075c-f223-46a3-b638-2a9f6e6d24fb" />

--- 
<img width="591" height="418" alt="Image" src="https://github.com/user-attachments/assets/5c6ae1c2-40f3-4d86-a10a-457c817016e9" />

### Expected Behavior

The expected behavior should cache the Sequential model were the symbolic input and output are respectively stored in model.inputs[0] and model.output[0]

### Version

1.3.3

### Environment

```markdown
- OS:linux
- Python version: 3.12
- Tensorflow version: 2.20.0
- Packages used version:
absl-py==2.3.1
accelerate==1.12.0
aiohappyeyeballs==2.6.1
aiohttp==3.13.3
aiosignal==1.4.0
annotated-types==0.7.0
antlr4-python3-runtime==4.9.3
anyio==4.12.0
astor==0.8.1
asttokens==3.0.1
astunparse==1.6.3
attrs==25.4.0
beartype==0.22.9
beautifulsoup4==4.14.3
bidict==0.23.1
bitsandbytes==0.49.1
bleach==6.3.0
certifi==2025.11.12
charset-normalizer==3.4.4
click==8.3.1
cloudpickle==3.1.2
comm==0.2.3
contourpy==1.3.3
cycler==0.12.1
debugpy==1.8.19
decorator==5.2.1
defusedxml==0.7.1
deprecated==1.3.1
diffusers==0.36.0
einops==0.8.2
executing==2.2.1
fastjsonschema==2.21.2
filelock==3.20.1
flatbuffers==25.9.23
fonttools==4.61.1
frozenlist==1.8.0
fsspec==2025.12.0
gast==0.7.0
ghp-import==2.1.0
gitdb==4.0.12
gitpython==3.1.46
google-pasta==0.2.0
grpcio==1.76.0
h11==0.16.0
h5py==3.15.1
[...]

PR #178, opened 2026-01-29
[Bug, Update]: tf.keras.Sequential error, uv, ruff, python 3.10, 3.11, 3.12, 3.13

quentinpossamai:
Trying to solve https://github.com/deel-ai/xplique/issues/177

As described in the issue, tf.keras.Sequential don't have the model.input and model.output defined even after a call.

These attributes are instead present in model.inputs[0] and model.output[0]. So I have added an elif condition in xplique.attributions.base.py `BlackBoxExplainer.__init__` to use the right attributes in case of a Sequential to build the caching of models.

TLDR:
Using id(model.inputs[0]) and id(model.outputs[0]) which contains the symbolic tensors (dtype and shape) of the input and output for a tf.keras.Sequential instead of id(model.input) and id(model.output).


Agustin-Picard (approved):
Great work! LGTM

quentinpossamai:
Changes:
- pre-commit hook to use ruff check and ruff format
- pyproject.toml replacing requirements.txt
- updated tox.ini to test on python 3.10, 3.11, 3.12, 3.13
- fix bugs when computing gradients from a tf.keras.Sequential, especially for tcav and gradcam *1
- updated .github/*.yml file for linting, publishing, docs building and testing
- updated contributing.md, integrating uv use


*1
In TCAV (very similary for GradCam):

`self.multi_head = tf.keras.Model(model.input, [target_layer.output, model.output])` was used to retrieve the gradient of the score w.r.t the taget_layer.output:
```
        with tf.GradientTape() as tape:
            tape.watch(inputs)

            activations, y_pred = self.multi_head(inputs)
            score = y_pred[:, label]

        gradients = tape.gradient(score, activations)
```
This failed when self.model is a `tf.keras.Sequential` (as model.output and model.input are never defined, even after a call), so I have `created watch_layer` and `end_watch_layer` to retrieve the activations:
```
        with tf.GradientTape(watch_accessed_variables=False) as tape:
            watch_layer(self.target_layer, tape)
            y_pred = self.model(inputs)
            activations = self.target_layer.result
            score = y_pred[:, label]

        gradients = tape.gradient(score, activations)
        end_watch_layer(self.target_layer)
```
We just need to keep the self.target_layer and do not need to build a second model (self.multi_head) now so I have deleted the multi_head attribute.

.github/workflows/python-lints.yml
```diff
@@ -1,26 +1,27 @@
...
 
 jobs:
   checks:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.7.14, 3.8, 3.9, "3.10"]
+        python-version: ["3.13"]
```
lucashervier: Why do we focus only on 3.13? If there is a good reason, then we don't need the matrix. But, in my opinion it would be good to keep on trying on versions we want to guarantee that the library is working.

.github/workflows/python-torch.yml
```diff
@@ -1,47 +1,65 @@
...
+          with open(os.environ["GITHUB_OUTPUT"], "a", encoding="utf-8") as f:
+              f.write(f"matrix={json.dumps(matrix)}\n")
+          PY
+
   checks:
+    needs: discover-tox-envs
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 4
+      fail-fast: false
```
lucashervier, line 52: Why did you add this option?

.github/workflows/python-torch.yml
lucashervier, line 1: For the workflow files in general, I will assume that you have done a functional test on your side!

docs/contributing.md
```diff
@@ -13,17 +13,17 @@ work is not in vain.
 - Clone the repo `git clone https://github.com/deel-ai/xplique.git`.
 - Go to your freshly downloaded repo `cd xplique`
 - Create a virtual environment and install the necessary dependencies for development `make prepare-dev && source xplique_dev_env/bin/activate`.
```
lucashervier: If we are using uv, maybe we should recommend using uv to create also the virtual env.

docs/contributing.md
```diff
@@ -51,8 +51,7 @@ Anyway, if you suceed you should see your virtual environment name in front of a
 ```
 - You can now install all necessary packages, with pip:
 ```
-pip install -r requirements.txt
-pip install -r requirements_dev.txt
+uv pip install -r pyproject.toml --group dev --group torch
```
lucashervier: Is it necessary to add torch dependency by default when developping? I think this should be a side command.

docs/contributing.md
```diff
@@ -79,15 +82,15 @@ So basically, if you do not succeed to use `make` just activate your virtual env
 `tox` on the otherhand will do the following:
 - run pytest on the tests folder with python 3.6, python 3.7 and python 3.8
 > Note: If you do not have those 3 interpreters the tests would be only performs with your current interpreter
-- run pylint on the xplique main files, also with python 3.6, python 3.7 and python 3.8
-> Note: It is possible that pylint throw false-positive errors. If the linting test failed please check first pylint output to point out the reasons.
+- run `ruff check` and `ruff format` on the xplique main files, also with python 3.6, python 3.7 and python 3.8
```
lucashervier: I think we need to update the python versions here

docs/contributing.md
lucashervier, line 1: Maybe it would be worth to be less verbose and focus on the "uv way" of installing (rather than giving all details for each distribution). Indeed, it is also one of the advantage of uv.

docs/index.md
lucashervier: To align with the README you also need to change this line with:

tests/attributions/test_grad_cam.py
```diff
@@ -30,11 +30,35 @@ def test_output_shape():
 
     for input_shape in input_shapes:
         samples, labels = generate_data(input_shape, nb_labels, 100)
+        model = tf.keras.Sequential(
```
lucashervier: I think those were for debug as we overwrite the model the line rightafter and the _generate_model method.

tests/attributions/test_grad_cam.py
```diff
@@ -30,11 +30,35 @@ def test_output_shape():
...
+            ]
+        )
+        model.compile(loss="categorical_crossentropy", optimizer="sgd")
+
         model = _generate_model(input_shape, nb_labels)
 
         method = GradCAM(model, -2)
         outputs = method.explain(samples, labels)
 
+        # debug
```
lucashervier: If debug is over we should clean that

tests/attributions/test_online_statistics.py
```diff
@@ -43,13 +43,17 @@ def test_online_square_mean():
 def test_online_variance():
     method = VarGrad(lambda x: None)
     for shape in [(1, 7, 4), (5, 2, 4, 4), (5, 7, 4, 4, 3)]:
-        samples = tf.reshape(tf.range(int(np.prod(shape)), dtype=tf.float32), shape)
+        # Force to go to tf.float64, as for the test last sample (5, 7, 4, 4, 3)
```
lucashervier: Add it in the comment here rather than on the code. We will have the same trace but in the PR rather than on the code ;)

tests/commons/test_segmentation_operator.py
```diff
@@ -47,63 +58,97 @@ def test_segmentation_operator():
...
+                [0, 0, 0],
+                [0, 0, 0],
+                [1.0, 1.0, 1.0],
+            ],
+        ]
+    ]
     target_2 = tf.transpose(tf.convert_to_tensor(target_2, tf.float32), perm=[0, 2, 3, 1])
 
-    scores = model(np.array(image)) * target_2
+    model(np.array(image)) * target_2
```
lucashervier, line 148: Is this line used??

xplique/attributions/object_detector.py
```diff
@@ -56,13 +53,15 @@ class BoundingBoxesExplainer(BlackBoxExplainer):
     """
 
     @deprecated(version="1.0.0", reason=OLD_OBJECT_DETECTION_DEPRECATION_MESSAGE)
```
lucashervier: We should even consider to remove that in upcoming versions

xplique/commons/operators_operations.py
```diff
@@ -293,4 +303,38 @@ def get_gradient_functions(
         batch_gradient = no_gradients_available
 
     return gradient, batch_gradient
-    
\ No newline at end of file
+
+
+def watch_layer(layer: tf.keras.layers.Layer, tape: tf.GradientTape) -> tf.keras.layers.Layer:
```
lucashervier, line 308: We need to have tests for the two functions below.

xplique/concepts/tcav.py
```diff
@@ -28,21 +28,16 @@ class Tcav:
...
-                 batch_size: Optional[int] = 64):
+    def __init__(
+        self, model: tf.keras.Model, target_layer: Union[str, int], batch_size: Optional[int] = 64
+    ):
         self.model = model
         self.batch_size = batch_size
 
         # configure model bottleneck
-        target_layer = find_layer(model, target_layer)
-        self.multi_head = tf.keras.Model(model.input, [target_layer.output, model.output])
```
lucashervier, line 40: I would like some explanations on why we do not need that anymore? What is the rationale behind this design change?

xplique/example_based/search_methods/proto_dash_search.py
```diff
@@ -28,20 +27,20 @@ class Optimizer():
...
+        initial_weights: Union[tf.Tensor, np.ndarray],
+        min_weight: float = 0,
+        max_weight: float = 10000,
     ):
         self.initial_weights = initial_weights
         self.min_weight = min_weight
         self.max_weight = max_weight
         self.bounds = [(min_weight, max_weight)] * initial_weights.shape[0]
-        self.objective_fn = lambda w, u, K: - (w @ u - 0.5 * w @ K @ w)
+        self.objective_fn = lambda w, u, K: -(w @ u - 0.5 * w @ K @ w)
```
lucashervier, line 39: In most codes you removed the lambda functions but not here. Is there any particular reason?

CONTRIBUTING.md
lucashervier, line 1: The comments I did on the other contributing file are relevant here too

Makefile
```diff
@@ -5,7 +5,7 @@ help:
 	@echo "make prepare-dev"
 	@echo "       create and prepare development environment, use only once"
 	@echo "make test"
-	@echo "       run tests and linting on py36, py37, py38"
+	@echo "       run tests and linting in current env"
```
lucashervier: in .venv/ would be better

package_python.py
```diff
@@ -0,0 +1,112 @@
...
+
+import requests
+from packaging.specifiers import SpecifierSet
+from packaging.version import InvalidVersion, Version
+
+
+def filename_implies_compat(filename: str, target: Version) -> bool:
+    t_major, t_minor = target.release[:2]
+
+    # # Look for py3 or py2.py3 in filename
```
lucashervier: Dead code

package_python.py
lucashervier, line 1: What is this file for and when do we use it? A lot of unnecessary commented code

pyproject.toml
```diff
@@ -0,0 +1,97 @@
...
+[[tool.bumpversion.files]]
+filename = "xplique/__init__.py"
+
+[tool.ruff]
+line-length = 100
+
+[tool.ruff.lint]
+select = ["E", "F"]
+ignore = [
+    "PLR0904",
```
lucashervier: Maybe we should have a short comment here telling what we are ignoring

pyproject.toml
```diff
@@ -0,0 +1,97 @@
...
+
+[tool.bumpversion]
+current_version = "1.4.2"
+commit = true
+tag = false
+
+[[tool.bumpversion.files]]
+filename = "pyproject.toml"
+
+[[tool.bumpversion.files]]
```
lucashervier, line 75: Don't we need this file too?

lucashervier (changes requested):
This is a very substantial PR — great work overall. There are many improvements and it’s clear a lot of effort went into it.

For future changes of this size, it might be beneficial to split them into several smaller PRs. In particular, quite a few diffs appear to be formatting-only changes. While those are useful, mixing them with functional updates makes the review a bit harder, as it becomes more difficult to quickly identify the core changes to focus on. I hope I didn’t miss anything important while going through it.

Overall, the direction looks very good 👍 I left a few questions and comments that I think should be clarified before merging, especially around the parts that modify method signatures or behavior.

I also assumed that the workflows were tested and that the existing tests were run successfully.

.github/workflows/deploy-docs.yml
```diff
@@ -15,22 +15,21 @@ jobs:
...
+        uses: astral-sh/setup-uv@v6
 
       - name: "Install Dependencies"
         run: |
-          python -m pip install --upgrade pip
-          pip install mkdocs mkdocs-material numkdoc mike
-          pip install -r requirements.txt
-          pip install torch
-          pip install -e .
+          uv pip install --system -r pyproject.toml --group dev --group torch -e .
```
Agustin-Picard: Is torch really needed for the docs?

.github/workflows/python-tests.yml
```diff
@@ -6,29 +6,47 @@ on:
...
-            - python-version: 3.7.14
-              tf-version: 2.2
-            - python-version: 3.8.12
-              tf-version: 2.5
-            - python-version: 3.9
-              tf-version: 2.8
-            - python-version: "3.10"
-              tf-version: 2.11
+          # Matches tox.ini (non-torch envs only)
+          - python-version: "3.10"
```
Agustin-Picard, line 15: We shouldn't forget to adjust the badges to the new compatible versions in the Readme and index of documentation

tests/attributions/test_online_statistics.py
Agustin-Picard: I think that's coming from tf with numpy 2 (or just np 2), btw

tests/commons/test_segmentation_operator.py
```diff
@@ -31,9 +42,9 @@ def default_methods(model, operator):
 
 
 def get_segmentation_model(input_shape=(20, 20, 1)):
-    model = tf.keras.Sequential([
-        tf.keras.layers.Input(input_shape),
-    ])
+    inp = tf.keras.layers.Input(input_shape)
```
Agustin-Picard, line 45: Is this coming from a newer version of tf that doesn't allow input to be the only layer?

tests/example_based/test_datasets_harmonization.py
```diff
@@ -118,7 +130,7 @@ def test_inputs_combinations():
     """
 
     tf_tensor = tf.reshape(tf.range(90, dtype=tf.float32), (10, 3, 3))
-    np_array = np.array(tf_tensor)
+    np.array(tf_tensor)
```
Agustin-Picard: unused conversion

tests/example_based/test_projections.py
```diff
@@ -178,7 +186,7 @@ def test_from_splitted_model():
     input_features = 10
     output_features = 3
     x_train = np.reshape(np.arange(0, nb_samples * input_features), (nb_samples, input_features))
-    tf_x_train = tf.convert_to_tensor(x_train, dtype=tf.float32)
+    tf.convert_to_tensor(x_train, dtype=tf.float32)  # tf_x_train
```
Agustin-Picard: unused code?

xplique/attributions/object_detector.py
Agustin-Picard: If we're considering bumping to a new major soon (and I think we should), maybe this is the right time to remove it

xplique/concepts/tcav.py
Agustin-Picard: I agree, there's no trace of this design choice in the PR description and I think it's important.

.pre-commit-config.yaml
```diff
@@ -1,10 +1,18 @@
...
     hooks:
     -   id: trailing-whitespace
     -   id: check-added-large-files
     -   id: check-ast
     -   id: check-docstring-first
     -   id: check-merge-conflict
     -   id: end-of-file-fixer
+-   repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.15.2
+    hooks:
```
Agustin-Picard, line 13: We should decide if we do both auto fix instead of fix violations but only check format

package_python.py
```diff
@@ -0,0 +1,112 @@
...
+                            release_ok = True
+                            break
+                    except Exception:
+                        pass
+            # Fallback to filename heuristics (wheels)
+            if filename_implies_compat(f.get("filename", ""), py_version):
+                release_ok = True
+                break
+
+        # # Fallback to project-level requires_python as no release-specific info found
```
Agustin-Picard: dead code

uv.lock
Agustin-Picard, line 1: Do we actually need to commit the uv.lock? For a library like Xplique that is tested with different versions of tf torch and python, I'm not sure this is relevant.

Agustin-Picard (changes requested):
Great work, you've implemented a lot of improvements that needed to be done (a long time ago for some of them)!

However, as @lucashervier said, maybe we should have split the PR into 2 or 3, as there were some important changes amidst all the formatting changes. No worries, but just keep it in mind for next time.

I see that you propose a `ruff check` and `ruff format` but only fix on PEP violations instead of doing both. We could technically fix both in the pre-commit (or only check in CI). What are your thoughts on this?

In this PR, you also propose changes in the way we perform GradCAM and TCAV, but they haven't been explained in the description. We should try to keep a trace of these changes somewhere as they seem out of the blue.

Finally, we should make sure that the docs build locally before merging. You wouldn't want to be like me and merge to find out you now need to make a second patch PR to fix the documentation ^^

.github/workflows/python-lints.yml
quentinpossamai: As this workflow just lints and formats the code (visual & beauty checks), the python version does not matter. I am removing the matrix.

.github/workflows/python-torch.yml
quentinpossamai: When set to 'false', failed jobs do not cancel running jobs. It is interesting to know which tox-envs will fail and which will succeed as sometimes it is a specific combination of packages that fails that I am not aware of.

.github/workflows/python-torch.yml
quentinpossamai: I have tested with GitHub workflows python-lints.yml, python-tests.yml and python-torch.yml. However, I have not tested deploy-docs.yml and python-publish.yml.

docs/contributing.md
quentinpossamai: Removing  torch

docs/contributing.md
quentinpossamai: Ok so I am dropping conda

tests/attributions/test_online_statistics.py
quentinpossamai: # Force to go to tf.float64, as for the test last sample (5, 7, 4, 4, 3)
        # At the last line of _get_online_statistic_final_value, (mean**2)[0, 1, 3, 0],
        # I got a loss of precision with 165.0**2 = 27224.998046875 and should be 27225.0

tests/commons/test_segmentation_operator.py
quentinpossamai: Before it was `scores = model(np.array(image)) * target_2`, but I got a linting error saying the variable `scores` was not used, so I just deleted the variable assignment while keeping the operation.

xplique/attributions/object_detector.py
quentinpossamai: I am removing `test/attributions/test_object_detector.py` and  `xplique/attributions/object_detector.py` then

xplique/concepts/tcav.py
quentinpossamai: `self.multi_head = tf.keras.Model(model.input, [target_layer.output, model.output])` was used to retrieve the gradient of the score w.r.t the `taget_layer.output`:
```
        with tf.GradientTape() as tape:
            tape.watch(inputs)

            activations, y_pred = self.multi_head(inputs)
            score = y_pred[:, label]

        gradients = tape.gradient(score, activations)
```
This failed when `self.model` is a `tf.keras.Sequential`, so I have created `watch_layer` and `end_watch_layer` to retrieve the activations:

```
        with tf.GradientTape(watch_accessed_variables=False) as tape:
            watch_layer(self.target_layer, tape)
            y_pred = self.model(inputs)
            activations = self.target_layer.result
            score = y_pred[:, label]

        gradients = tape.gradient(score, activations)
        end_watch_layer(self.target_layer)
```

We just need to keep the `self.target_layer` and do not need to build a second model (`self.multi_head`) now so I deleted the `multi_head` attribute.

I also have updated the PR's description

xplique/example_based/search_methods/proto_dash_search.py
quentinpossamai: Here `ruff` does not raise an issue when using `lambda` because it is an attribute (and not a variable) assignment. You can also use `lambda` to be passed to arguments in a function call.

package_python.py
quentinpossamai: Removing it, it was used to discover pipy packages I now use `pip index` and learned that `uv pip index` does not exist.

pyproject.toml
quentinpossamai: "PLR0904",  # https://docs.astral.sh/ruff/rules/too-many-public-methods/
    "PLR0914",  # https://docs.astral.sh/ruff/rules/too-many-locals/
    "PLR0917",  # https://docs.astral.sh/ruff/rules/too-many-positional-arguments/
    
    They were ignored with the old setup.cfg

pyproject.toml
quentinpossamai: So `pyproject.toml`, `uv.lock` are now replacing `requirements.txt`, `requirements.dev` and `setup.cfg` as it is the new standard with `uv`.

.github/workflows/deploy-docs.yml
quentinpossamai: Removing torch
Sorry for the lack of clarity, I will make several PR next time
I have updated .pre-commit-config.yaml regarding ruff check and format
I have updated the PR's description regarding TCAV and GradCam

I have now tested to build the project and the doc

.github/workflows/python-tests.yml
quentinpossamai: Badges have now been updated on the ReadMe and index

docs/contributing.md
quentinpossamai: Updating both contributing.md

tests/commons/test_segmentation_operator.py
quentinpossamai: Only one Input layer in a Sequential does not provide the attributes .inputs and .outputs used by the mother class BlackBoxExplainer

.pre-commit-config.yaml
quentinpossamai: I see the inconsistency, I propose to auto fix then, the changes are not staged and committed automatically so the dev will be able to see them

package_python.py
quentinpossamai: Deleting

uv.lock
quentinpossamai: It is used for dev only if I understood correctly to sync dev env, uv recommend tracking it. I am dropping it, it is not *needed*.

tests/attributions/test_object_detector.py
Agustin-Picard, line 1: Why did this test file get removed?

xplique/attributions/object_detector.py
Agustin-Picard, line 1: Why did this test get removed?

tests/attributions/test_object_detector.py
Agustin-Picard: I just saw the message in another comment. I'm resolving.

xplique/attributions/object_detector.py
Agustin-Picard: Same.

Agustin-Picard (approved):
Looks good to me. The requested changes have been made as far as I'm concerned.

lucashervier (approved):
LGTM ;)

PR #184, opened 2026-08-26
fix: attach tabular colorbar to plot axes

### LINKED ISSUE #159: [Bug]: - Tabular plot does not support numpy 1.26.3
### Module

Attributions Methods

### Current Behavior

When calling the tabular plot function for attributions with numpy version 1.2.6, an error is raised while it it not the case for earlier versions of numpy.

### Expected Behavior

The library should support the latest versions of numpy.

### Version

1.3.3

### Environment

```markdown
- numpy 1.26.3
```

### Relevant log output

```shell
At `xplique.plots.tabular.py` line 49: `colorbar = plt.colorbar(mappa, ticks=[0, 1])`.

The following error is raised:
`ValueError: Unable to determine Axes to steal space for Colorbar. Either provide the *cax* argument to use the Axes for the Colorbar, provide the *ax* argument to steal space from it or add *mappable* to an Axes`.
```

### To Reproduce

Launch the plot function with latest version of numpy.

rashidrao-pk:
## Description

Fixes #159.

`summary_plot_tabular` created a standalone `ScalarMappable` and passed it to `plt.colorbar()` without specifying the Axes from which the colorbar should take space. Recent Matplotlib versions therefore raise:

> ValueError: Unable to determine Axes to steal space for Colorbar.

This change passes the active tabular plot Axes explicitly to `plt.colorbar(..., ax=ax)`.

## Changes

* Update `_add_colorbar` to accept the target Axes.
* Attach the colorbar explicitly to the tabular summary plot.
* Add a regression test covering colorbar creation.

## Validation

* `MPLBACKEND=Agg pytest -q tests/plots/test_tabular_plots.py` — 1 passed
* `MPLBACKEND=Agg pytest -q tests/plots` — 16 passed
* `ruff check xplique/plots/tabular.py tests/plots/test_tabular_plots.py` — passed
* `ruff format --check xplique/plots/tabular.py tests/plots/test_tabular_plots.py` — passed

### DIFF
--- a/tests/plots/test_tabular_plots.py
+++ b/tests/plots/test_tabular_plots.py
@@ -1,0 +1,19 @@
+"""Tests for tabular plotting utilities."""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from xplique.plots import summary_plot_tabular
+
+
+def test_summary_plot_tabular_adds_colorbar():
+    """The summary plot attaches its colorbar to the plot axes."""
+    explanations = np.array([[0.2, -0.1], [0.4, 0.3]])
+    features_values = np.array([[0.1, 0.8], [0.9, 0.2]])
+
+    summary_plot_tabular(explanations, features_values)
+
+    figure = plt.gcf()
+    assert len(figure.axes) == 2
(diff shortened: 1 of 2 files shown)


AntoninPoche (approved):
Looks good to me!

Thank you very much for your contribution!

PR #185, opened 2026-08-27
Bump version: 2.0.0 → 2.0.1 + new regex

AntoninPoche:
Initially, I had the wrong order and tried to bump the version after the merge, which forced me to create a new PR.

However, through the PR, I discovered that the bump version was buggy: it searched for the version number and replaced its occurrences with the new version number. But this could also change dependency versions.

So I updated the regex and ran the bump version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "xplique"
-version = "2.0.0"
+version = "2.0.1"
 description = "Explanations toolbox for TensorFlow 2"
 readme = "README.md"
 requires-python = ">=3.10,<3.14"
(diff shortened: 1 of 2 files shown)


pyproject.toml
AntoninPoche, line 1: The bump version script should only update the lib version, not dependencies.

Agustin-Picard:
Thanks for catching this Antonin! I think I've found a slightly simpler way: we do not need to specify the file for the version if it's already stored in `pyproject.toml` (as per https://callowayproject.github.io/bump-my-version/howtos/multiple-replacements/).

As we've discussed, I'll take over and squash the commits for a cleaner git history. I'll let you review when I'm done.

Thanks again!

AntoninPoche (commented):
LGTM

Agustin-Picard (approved):

PR #186, opened 2026-08-31
Add a ConceptLocalizer via Black-box Attributions

Agustin-Picard:
This PR adds **input-to-concept localization** to `HolisticCraft`.

The new API answers:

> **Which input regions drive the activation of concept `k`?**

It does so by exposing the fitted latent extractor and factorizer as a model that returns one **concept activation score** per concept, then applying a standard Xplique **black-box attribution method** to the selected score.

This complements the existing visualization of latent concept activation maps; it does not replace it.

The implementation is currently validated with `Rise` and `SobolAttributionMethod` and is designed to work with compatible black-box attribution methods.

## Motivation

`HolisticCraft` already produces concept coefficients `coeffs_u`.

When the latent representation is spatial, these coefficients form **concept activation maps**:

```text
U_k(x)
```

which can be resized and overlaid on the input.

With the default spatial mean reducer, the concept score is
`s_k(x) = mean_{i,j} U_k(x)_{i,j}`. Resizing `U_k(x)` is therefore a CAM-style
map for the **concept score**: it displays the spatial map underlying that score
directly, without gradients or class-prediction weights. This works well for CNN
architectures, but may not render well for very specific deep architectures, 
large receptive fields, attention-based models, multi-scale representations, 
or other cases where a latent position does not correspond
directly to a local input region.

This PR adds the complementary input-to-concept path. It computes a concept
score, then a black-box explainer perturbs input regions to attribute that score
to the input:

```text
input
  -> latent activations
  -> factorizer.encode(...)
  -> concept activation map U_k(x)
  -> reducer R
  -> concept activation score s_k(x)
  -> black-box input attribution
  -> concept localization map
```

## Terminology

The implementation and documentation use four distinct quantities:

| Quantity | Definition | Question answered |
|---|---|---|
| **Concept activation map** | Spatial factorization coefficients `U_k(x)`, stored in `coeffs_u`; CAM-style map for the spatial mean concept score | Where and how strongly is concept `k` activated in latent space? |
| **Concept activation score** | Scalar reduction `s_k(x) = R(U_k(x))` | How strongly is concept `k` activated overall? |
| **Concept importance** | Attribution of the task prediction to concept activations | How much do concept activations contribute to the task prediction? |
| **Concept localization map** | Attribution of a concept activation score to the input | Which input regions drive the concept activation score? |

For perturbation-based methods, the two attribution paths vary different quantities:

```text
Concept importance (Sobol): perturb latent concept activations
    -> measure the task prediction -> attribute it to concepts

Concept localization (RISE/Sobol): perturb input regions
    -> measure concept activation score s_k(x) -> attribute it to input regions
```

Concept importance also supports gradient-based methods; the first path
describes Sobol-based importance.

For CNN activations with shape `(N, H', W', C)`, factorization produces `coeffs_u` with shape `(N, H', W', K)`, so each `U_k(x)` is spatial.

Spatial ViT tokens can similarly form a concept activation map when mapped back to their patch grid.

For global representations with shape `(N, K)`, there is no spatial concept activation map; the coefficient for each concept is already a global activation score.

### `make_concept_localizer()`

```python
localizer = craft.make_concept_localizer(
    concept_reducer="mean",
)
```

The returned callable maps a batch of input images to:

```text
(batch_size, number_of_concepts)
```

concept activation scores.

Built-in reducers are:

```text
"mean"
"sum"
"max"
```

A custom callable can also be provided as long as it returns one scalar per input and concept.

The localizer preserves signed concept coefficients. Users who intentionally want magnitude-based scores can provide an appropriate custom reducer.

The concrete TensorFlow/PyTorch localizer implementations are private implementation details; the public entry point is `HolisticCraft.make_concept_localizer()`.

### `attribute_concepts_to_inputs()`

```python
concept_maps = craft.attribute_concepts_to_inputs(
    images,
    partial_explainer=partial_explainer,
    concept_ids=[0, 3, 7],
    concept_reducer="mean",
)
```

This method:

1. validates the requested concepts;
2. creates the concept-score localizer;
3. instantiates the configured black-box explainer once;
4. uses one-hot concept targets;
5. attributes each requested concept activation score to the input.

For image inputs, the result has shape:

```text
(N, H, W, number_of_concepts)
```

Channel `k` always corresponds to concept `k`.

If only a subset of concepts is requested, uncomputed channels are filled entirely with `NaN`, so they cannot be confused with valid zero attribution.

TensorFlow callers use native channel-last inputs. PyTorch callers pass native channel-first tensors to the high-level method; the framework wrapper handles the layout conversion required by the attribution stack.

## Why localization is black-box only

Input-to-concept localization evaluates the fitted factorizer on perturbed inputs through:

```python
factorizer.encode(...)
```

This path is not guaranteed to be differentiable.

For this reason, `attribute_concepts_to_inputs()` currently supports black-box attribution methods only.

Gradient-based methods remain supported for the separate **concept importance** path, where concept activations are decoded toward the task prediction.

No task `operator` is required for concept localization because one-hot targets directly select a concept activation score.

If a non-`None` `operator` is supplied in the `PartialExplainer`, it is ignored with a warning.

Factorizers used for localization must support out-of-sample `encode()`. A non-inductive factorizer raises an explicit runtime error when localization is requested.

### Sobol

```python
from xplique.attributions import SobolAttributionMethod
from xplique.concepts import PartialExplainer

sobol = PartialExplainer(
    SobolAttributionMethod,
    grid_size=8,
    nb_design=32,
    perturbation_function="inpainting",
)

concept_maps = craft.attribute_concepts_to_inputs(
    model_images,
    partial_explainer=sobol,
    concept_ids=[0, 3, 7],
    concept_reducer="mean",
)

craft.display_images_per_concept(
    display_images,
    concept_maps=concept_maps,
    order=[0, 3, 7],
)
```

### RISE

RISE uses the same public localization API; only the `PartialExplainer` configuration changes:

```python
from xplique.attributions import Rise
from xplique.concepts import PartialExplainer

rise = PartialExplainer(
    Rise,
    nb_samples=2000,
    grid_size=7,
    preservation_probability=0.5,
    mask_value=0.0,
)

concept_maps = craft.attribute_concepts_to_inputs(
    model_images,
    partial_explainer=rise,
    concept_ids=[0, 3, 7],
)
```

## Visualization integration

The existing display API now supports both interpretation modes.

### Concept activation map visualization

Existing behavior remains available through `coeffs_u`:

```python
coeffs_u = craft.transform(model_images)

craft.display_images_per_concept(
    display_images,
    coeffs_u=coeffs_u,
)
```

For spatial coefficients and the default mean score, this is a CAM-style
visualization of the concept score in latent space. No input perturbations are
used to produce this map.

### Concept localization maps

Input attribution maps can instead be provided with:

```python
craft.display_images_per_concept(
    display_images,
    concept_maps=concept_maps,
)
```

`display_top_images_per_concept()` also accepts `concept_maps`, with an important semantic distinction:

- **`coeffs_u` determines which images rank highest for a concept**;
- **`concept_maps` determines the localization overlay**.

Concept localization magnitude therefore does not replace concept activation as the ranking signal.

If a localization map already matches the input resolution, `display_concept_heatmap()` avoids an unnecessary resize.

Signed localization maps are currently displayed by absolute magnitude. Neither concept activation maps nor concept localization maps should be interpreted as segmentation masks.

## Backward compatibility

The feature is opt-in.

Existing usage of:

```text
display_images_per_concept(...)
display_top_images_per_concept(...)
estimate_importance(...)
compute_explanation_per_concept(...)
get_topk_images_per_concept(...)
ConceptDecoder
```

retains its previous behavior.

No existing call triggers black-box input attribution unless `attribute_concepts_to_inputs()` is explicitly requested.

## Tests

The PR includes coverage for:

- concept activation score reducers: `mean`, `sum`, `max`, and custom callables;
- spatial, token, and global concept coefficient layouts;
- TensorFlow/PyTorch concept-score parity;
- TensorFlow and PyTorch black-box localization smoke tests;
- concept selection and one-hot target construction;
- output shape and finite-value validation;
- partial localization with `NaN` channels for uncomputed concepts;
- invalid/duplicate concept IDs and invalid image batches;
- ignored explicit `operator` behavior;
- black-box-only validation;
- non-inductive factorizer failures;
- localization-map display behavior;
- preservation of coefficient-based top-image ranking;
- a semantic perturbation test showing that localization follows the concept's spatial dependency.

Cross-framework parity tests and framework-agnostic localization regressions are kept separate from the PyTorch-specific classification tests to keep the test responsibilities explicit.

## Documentation

The HolisticCraft documentation now:

- defines concept activation maps, concept activation scores, concept importance, and concept localization maps consistently;
- explains when a concept activation map is a CAM-style visualization of the concept score;
- includes a compact end-to-end workflow tied to the actual API;
- explains the distinction between latent activation-map visualization and input attribution;
- explains CNN, spatial-token, and global coefficient layouts;
- documents the black-box-only restriction and factorizer `encode()` requirement;
- provides Sobol and RISE localization examples;
- documents the interaction between `coeffs_u`, `concept_maps`, and top-image ranking.

Updated notebook:

https://drive.google.com/file/d/19eg2uU95A_LW1hQ1ENrhGI41lLK35UHu/view?usp=sharing

### DIFF
--- a/docs/api/concepts/holistic_craft.md
+++ b/docs/api/concepts/holistic_craft.md
@@ -64,8 +64,10 @@ Holistic CRAFT follows the same core principle as CRAFT but operates on full ima
     Third-party NMF implementations may not have this limitation
     (e.g., the Semi-NMF from the Overcomplete library).
 
-3. **Estimate Importance**: Use any attribution methods available in Xplique (gradient-based, perturbation-based) to rank concept importance
-4. **Visualize**: Generate concept heatmaps overlaid on images to show "what" and "where"
+3. **Measure Concept Activation**: Use concept coefficients to measure how strongly each concept is present and rank representative images
+4. **Estimate Concept Importance**: Attribute task predictions to concept coefficients to measure how much each concept contributes to the model output
+5. **Visualize Coefficients**: Resize latent coefficient maps and overlay them on input images
+6. **Localize Concepts**: Optionally use a black-box attribution method to identify which input regions drive a selected concept score
 
 Like regular CRAFT, Holistic CRAFT requires splitting the model into two parts: $(g, h)$ such that $f(x) = (g \cdot h)(x)$. The model $g$ maps input to latent space (activation maps), and $h$ maps latent space to predictions. Concepts are extracted from these activation maps in latent space.
 
@@ -76,6 +78,37 @@ This split is implemented through three abstractions:
 - **`LatentExtractor`**: Wraps both $g$ (`input_to_latent_model`) and $h$ (`latent_to_logit_model`). It orchestrates the full forward pass, batching, device management, and output formatting. The `TorchLatentExtractor` and `TfLatentExtractor` subclasses provide framework-specific implementations.
 
 - **`LatentExtractorBuilder`**: A factory that constructs a `LatentExtractor` for a specific model architecture. It handles all the architecture-specific wiring (defining how to split the model, which layer to extract from, and how to format outputs) so that the rest of the CRAFT pipeline remains model-agnostic.
+
+## Concept Activation, Importance, and Localization
+
+Holistic CRAFT exposes three related quantities that answer different questions:
+
+| Quantity | Definition | Question answered |
+|---|---|---|
+| **Concept activation** | The spatial coefficient map $U_k(x)$, or a reduction of it | How strongly is concept $k$ present? |
+| **Concept importance** | Attribution of the final task prediction to concept $k$ | How much does concept $k$ contribute to the prediction? |
+| **Concept localization** | Attribution of the concept score $s_k(x) = R(U_k(x))$ to the input | Which input regions drive the concept score? |
+
+Concept importance and concept localization follow opposite attribution directions:
+
+```text
+concept coefficients -> final prediction -> concept importance
+
+input image -> concept scores -> input attribution -> concept localization
+```
+
+A coefficient heatmap is a latent-space map resized to the input resolution. A localization
+map is an input-space attribution produced by perturbing the input and observing changes in a
+selected concept score. The maps can therefore differ, especially when the latent
+representation is coarse or its positions have large receptive fields.
+
+Top-image ranking always uses concept coefficients because they represent concept presence.
+Attribution-map magnitude is not a replacement for concept activation.
+
+!!!note
+    Coefficient and localization maps are explanations, not segmentation masks. Black-box
+    localization is not automatically more correct than coefficient visualization; it answers
+    the more specific question of which input perturbations change a selected concept score.
 
 
 ## Example
@@ -90,8 +123,8 @@ from xplique_adapters.concepts.torch.latent_data_retinanet import RetinanetExtra
 # Build a latent extractor that splits the model into g(.) and h(.)
 # This provides the input_to_latent (g) and latent_to_logit (h) functions
 latent_extractor = RetinanetExtractorBuilder.build(
-    model, 
-    device="cuda", 
+    model,
+    device="cuda",
     nb_classes=91,
     extraction_location='resnet', # Choose 'resnet' or 'fpn'
     extraction_layer=-1  # Extract from last ResNet feature layer
@@ -162,14 +195,167 @@ explanation_vargrad = craft.compute_explanation_per_concept(
     confidence=0.3,
 )
 
-# Reduce the spatial dimension of the explanation 
+# Reduce the spatial dimension of the explanation
 # to compute the final concepts importances
 importances_vargrad = craft.reduce_to_importance(
     explanation=explanation_vargrad,
 )
 ```
 
-### Using a Different NMF Factorizer
+## Localizing Concepts with Black-Box Attribution
+
+`ConceptLocalizer` exposes the fitted encoder and factorizer as a callable that returns one
+scalar score per learned concept. `compute_concept_attributions()` builds this callable,
+constructs one-hot concept targets, and applies a compatible black-box explainer to each
+requested concept.
+
+For an input batch with shape `(N, H, W, C)`, the returned maps have shape
+`(N, H, W, number_of_concepts)`. Channel `k` always corresponds to concept `k`. When only a
+subset is requested, uncomputed channels are filled entirely with `NaN` so that they cannot
+be confused with valid zero attributions.
+
+TensorFlow callers pass channel-last images. PyTorch callers pass their native channel-first
+`(N, C, H, W)` images to the same high-level method; Xplique handles the layout conversion
+for the wrapped localizer.
+
+### Localizing Concepts with RISE
+
+```python
+from xplique.attributions import Rise
+from xplique.concepts import PartialExplainer
+
+rise = PartialExplainer(
+    Rise,
+    nb_samples=2000,
+    grid_size=7,
+    preservation_probability=0.5,
+    mask_value=0.0,
+)
+
+rise_maps = craft.compute_concept_attributions(
+    images,
+    partial_explainer=rise,
+    concept_ids=[0, 3, 7],
+    concept_reducer="mean",
+)
+
+craft.display_images_per_concept(
+    display_images,
+    concept_maps=rise_maps,
+    order=[0, 3, 7],
+)
+```
+
+`concept_reducer="mean"` reduces every spatial coefficient map to the scalar score that RISE
+attributes to the input. It is the default and is consistent with the mean coefficient
+activation used to rank representative images. The localizer preserves signed scores; a
+custom callable reducer can be used when concept magnitude is intended instead.
+
+Do not pass a task `operator` to `PartialExplainer` for concept localization. One-hot targets
+are created internally to select concept scores directly.
+
+### Localizing Concepts with Sobol
+
+The same API works with `SobolAttributionMethod`:
+
+```python
+from xplique.attributions import SobolAttributionMethod
+
+sobol = PartialExplainer(
+    SobolAttributionMethod,
+    grid_size=8,
+    nb_design=32,
+    perturbation_function="inpainting",
+)
+
+sobol_maps = craft.compute_concept_attributions(
+    images,
+    partial_explainer=sobol,
+    concept_ids=[0, 3, 7],
+)
+
+craft.display_images_per_concept(
+    display_images,
+    concept_maps=sobol_maps,
+    order=[0, 3, 7],
+)
+```
+
+`nb_design` must be a nonzero power of two. Keep Sobol's default `nb_channels=1`: concept
+localization computes a separate single-channel attribution map for each selected concept.
+This differs from Sobol-based concept importance, which attributes the final task prediction
+to concept coefficients.
+
+### Coefficient Maps and Localization Maps
+
+Use `coeffs_u` to display latent coefficient maps:
+
+```python
+craft.display_images_per_concept(
+    display_images,
+    coeffs_u=coefficients,
+    order=selected_concepts,
+)
+```
+
+Use `concept_maps` to display input-space localization maps:
+
+```python
+craft.display_images_per_concept(
+    display_images,
+    concept_maps=rise_maps,
+    order=selected_concepts,
+)
+```
+
+When displaying the top images, coefficients still determine the ranking and localization
+maps only determine the overlay:
+
+```python
+craft.display_top_images_per_concept(
+    display_images,
+    coeffs_u=coefficients,
+    concept_maps=rise_maps,
+    order=selected_concepts,
+    topk=3,
+)
+```
+
+If `order` is omitted, display methods show only the concept channels that were computed.
+Explicitly requesting an uncomputed `NaN` channel raises a `ValueError`. Signed localization
+maps are displayed by absolute magnitude; attribution direction is not represented by the
+current renderer.
+
+!!!warning "Computational cost"
+    Black-box localization evaluates the encoder and `factorizer.encode()` for many perturbed
+    inputs and runs a separate attribution pass for every selected concept. Estimate concept
+    importance first, then localize only a small number of important concepts. Reduce the
+    number of images, RISE samples, Sobol designs, or grid resolution for exploratory runs.
+
+!!!warning "Perturbations use preprocessed inputs"
+    Perturbation parameters operate in the model's preprocessed input space. For example,
+    `mask_value=0.0` is neutral or black only if zero has that meaning after preprocessing.
+    With standard ImageNet normalization, zero generally corresponds to the dataset mean
+    rather than a raw black pixel. Sobol inpainting and blurring must be interpreted relative
+    to the same model-ready representation.
+
+!!!warning "Factorizer compatibility"
+    Localization evaluates the fitted factorizer on unseen, perturbed activations. The
+    factorizer must therefore support out-of-sample `encode()`. Activations should not be
+    clipped merely to satisfy a factorizer because that would change the function being
+    explained.
+
+!!!warning "Black-box scope"
+    Input-to-concept localization currently supports black-box attribution methods only.
+    Ordinary `factorizer.encode()` is not guaranteed to be differentiable, so white-box
+    explainers such as Gradient Input or Integrated Gradients are rejected.
+
+!!!warning "Interpretation"
+    Localization maps measure sensitivity of a concept score to input perturbations. They are
+    not object boundaries or segmentation labels. RISE and Sobol can also produce different
+    maps because they use different perturbation and aggregation strategies.
+
+## Using a Different NMF Factorizer
 
 By default, the standard Sklearn NMF is used to factorize the concepts.
 But other types of factorizers are supported, such as the ones provided
@@ -220,7 +406,7 @@ class CustomLatentData(LatentData):
     def get_activations(self, as_numpy: bool = True, keep_gradients: bool = False):
         """Extract activations from the specified layer."""
         activations = self.fpn_outs[self.extraction_layer]
-        
+
         if not keep_gradients:
             activations = activations.detach()
 
@@ -264,7 +450,7 @@ class CustomExtractorBuilder(LatentExtractorBuilder):
         extraction_layer: int = -1,
         batch_size: int = 1
     ) -> TorchLatentExtractor:
-        
+
         # Define g(.) function: input → latent activations
         def g(self, x):
             # Example: extract from backbone/feature pyramid
@@ -297,7 +483,7 @@ class CustomExtractorBuilder(LatentExtractorBuilder):
             batch_size=batch_size,
             device=device
         )
-        
+
         # Store extraction layer for later use
         latent_extractor.extraction_layer = extraction_layer
         return latent_extractor
@@ -343,6 +529,8 @@ craft.display_images_per_concept(input_images[:5])
 {{xplique.concepts.holistic_craft.HolisticCraft}}
 
 {{xplique.concepts.holistic_craft.PartialExplainer}}
+
+{{xplique.concepts.holistic_craft.ConceptLocalizer}}
 
 ## References
 
--- a/tests/concepts/test_holistic_craft_classification_torch.py
+++ b/tests/concepts/test_holistic_craft_classification_torch.py
@@ -3,6 +3,7 @@
 
 import numpy as np
 import pytest
+import tensorflow as tf
 import torch
 import torchvision.transforms as T
 from PIL import Image
@@ -13,11 +14,34 @@ from xplique.attributions import Rise, Saliency, SobolAttributionMethod
 from xplique.attributions.gradient_input import GradientInput
 from xplique.concepts import HolisticCraftTorch as Craft
 from xplique.concepts.holistic_craft import PartialExplainer
-from xplique.concepts.torch.holistic_craft import ConceptLocalizerTorch
-from xplique.concepts.torch.layered_model_latent_extractor import LayeredModelExtractorBuilder
+from xplique.concepts.tf.holistic_craft import HolisticCraftTf
+from xplique.concepts.tf.latent_extractor import TfLatentExtractor
+from xplique.concepts.tf.layered_model_latent_extractor import LayeredLatentData as TfLatentData
+from xplique.concepts.torch.latent_extractor import TorchLatentExtractor
+from xplique.concepts.torch.layered_model_latent_extractor import (
+    LayeredLatentData as TorchLatentData,
+)
+from xplique.concepts.torch.layered_model_latent_extractor import (
+    LayeredModelExtractorBuilder,
+)
 from xplique.utils_functions.classification.torch.classifier_tensor import TorchClassifierTensor
 from xplique.utils_functions.common.torch.gradients_check import check_model_gradients
 from xplique.wrappers import TorchWrapper
+
+
+class _IdentityFactorizer:
+    is_fitted = False
+    requires_positive_activations = False
+
+    def fit(self, activations):
+        self.is_fitted = True
+        return np.eye(2, dtype=np.float32), np.asarray(activations, dtype=np.float32)
+
+    def encode(self, activations):
+        return np.asarray(activations, dtype=np.float32)
+
+    def encode_differentiable(self, activations):
+        return activations
 
 
 def test_classifier_tensor_targets_class_and_preserves_batch_shape():
@@ -135,17 +159,15 @@ def test_latent_extractor(image_data, latent_extractor_data):
 
 def test_layered_latent_data_getitem(device_param):
     """Test that LayeredLatentData supports integer and slice indexing."""
-    from xplique.concepts.torch.layered_model_latent_extractor import LayeredLatentData
-
     activations = torch.randn(4, 16, 7, 7).to(device_param)
-    latent_data = LayeredLatentData(activations)
+    latent_data = TorchLatentData(activations)
 
     sliced = latent_data[1:3]
-    assert isinstance(sliced, LayeredLatentData)
+    assert isinstance(sliced, TorchLatentData)
     assert sliced.activations.shape[0] == 2
 
     item = latent_data[0]
-    assert isinstance(item, LayeredLatentData)
+    assert isinstance(item, TorchLatentData)
 
 
 def test_latent_extractor_gradients(image_data, latent_extractor_data):
@@ -208,6 +230,32 @@ def craft_data(image_data, latent_extractor_data, device_param):
     craft.fit(input_tensor)
 
     return craft
+
+
+def _make_tiny_torch_craft(device):
+    values = np.arange(2 * 4 * 4 * 2, dtype=np.float32).reshape(2, 4, 4, 2)
+    images = torch.from_numpy(values.transpose(0, 3, 1, 2)).to(device)
+    extractor = TorchLatentExtractor(
+        model=torch.nn.Identity(),
+        input_to_latent_model=lambda inputs: TorchLatentData(inputs),
+        latent_to_logit_model=lambda latent_data: latent_data.activations,
+        device=str(device),
+        batch_size=2,
+    )
+    craft = Craft(
+        extractor,
+        number_of_concepts=2,
+        device=str(device),
+        factorizer=_IdentityFactorizer(),
+    )
+    craft.fit(images)
+    return craft, images, values
+
+
+@pytest.fixture
+def tiny_craft_data(device_param):
+    """Create a deterministic identity CRAFT pipeline for localization tests."""
+    return _make_tiny_torch_craft(device_param)
 
 
 def test_craft_reencode(image_data, craft_data):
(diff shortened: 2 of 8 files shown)


docs/api/concepts/holistic_craft.md
```diff
@@ -64,8 +64,10 @@ Holistic CRAFT follows the same core principle as CRAFT but operates on full ima
     Third-party NMF implementations may not have this limitation
     (e.g., the Semi-NMF from the Overcomplete library).
 
-3. **Estimate Importance**: Use any attribution methods available in Xplique (gradient-based, perturbation-based) to rank concept importance
-4. **Visualize**: Generate concept heatmaps overlaid on images to show "what" and "where"
+3. **Measure Concept Activation**: Use concept coefficients to measure how strongly each concept is present and rank representative images
+4. **Estimate Concept Importance**: Attribute task predictions to concept coefficients to measure how much each concept contributes to the model output
+5. **Visualize Coefficients**: Resize latent coefficient maps and overlay them on input images
+6. **Localize Concepts**: Optionally use a black-box attribution method to identify which input regions drive a selected concept score
```
AntoninPoche: When giving steps like this. I find it useful to have a minimal code snippet just after where these numbers correspond to comments. It concretizes the saying, linking it to method names.

docs/api/concepts/holistic_craft.md
```diff
@@ -77,6 +79,37 @@ This split is implemented through three abstractions:
...
+| **Concept activation** | The spatial coefficient map $U_k(x)$, or a reduction of it | How strongly is concept $k$ present? |
+| **Concept importance** | Attribution of the final task prediction to concept $k$ | How much does concept $k$ contribute to the prediction? |
+| **Concept localization** | Attribution of the concept score $s_k(x) = R(U_k(x))$ to the input | Which input regions drive the concept score? |
+
+Concept importance and concept localization follow opposite attribution directions:
+
+```text
+concept coefficients -> final prediction -> concept importance
+
+input image -> concept scores -> input attribution -> concept localization
```
AntoninPoche: You use different names for concept coefficients and scores; it should be harmonized with the above table (activations).

docs/api/concepts/holistic_craft.md
```diff
@@ -77,6 +79,37 @@ This split is implemented through three abstractions:
...
+```
+
+A coefficient heatmap is a latent-space map resized to the input resolution. A localization
```
AntoninPoche: "Coefficient heatmap" was not defined previously.

Ensure terms consistency, do not use synonyms and reformulations. It is better to copy it many times.

docs/api/concepts/holistic_craft.md
AntoninPoche: What is the difference between input attribution and concept localization?

docs/api/concepts/holistic_craft.md
```diff
@@ -77,6 +79,37 @@ This split is implemented through three abstractions:
...
+map is an input-space attribution produced by perturbing the input and observing changes in a
+selected concept score. The maps can therefore differ, especially when the latent
+representation is coarse or its positions have large receptive fields.
+
+Top-image ranking always uses concept coefficients because they represent concept presence.
+Attribution-map magnitude is not a replacement for concept activation.
+
+!!!note
+    Coefficient and localization maps are explanations, not segmentation masks. Black-box
```
AntoninPoche: What is a coefficient map? I see a meaning to it only for ViTs or LLMs. For CNNs, Can we also make this map?

docs/api/concepts/holistic_craft.md
```diff
@@ -162,14 +195,167 @@ explanation_vargrad = craft.compute_explanation_per_concept(
...
+
+rise = PartialExplainer(
+    Rise,
+    nb_samples=2000,
+    grid_size=7,
+    preservation_probability=0.5,
+    mask_value=0.0,
+)
+
+rise_maps = craft.compute_concept_attributions(
```
AntoninPoche: unclear name.
is it inputs-to-concepts or concepts-to-outputs?

Proposition: `.attribute_concepts_to_inputs`

docs/api/concepts/holistic_craft.md
```diff
@@ -162,14 +195,167 @@ explanation_vargrad = craft.compute_explanation_per_concept(
...
+)
+```
+
+`concept_reducer="mean"` reduces every spatial coefficient map to the scalar score that RISE
+attributes to the input. It is the default and is consistent with the mean coefficient
+activation used to rank representative images. The localizer preserves signed scores; a
+custom callable reducer can be used when concept magnitude is intended instead.
+
+Do not pass a task `operator` to `PartialExplainer` for concept localization. One-hot targets
+are created internally to select concept scores directly.
```
AntoninPoche: Having an assertion or just ignoring it would make more sens.

Do not hesitate to use code comments for such comments:

```python
rise = PartialExplainer(
    Rise,
    # no need to specify the operator
    nb_samples=2000,
    grid_size=7,
    preservation_probability=0.5,
    mask_value=0.0,
)
```

docs/api/concepts/holistic_craft.md
```diff
@@ -162,14 +195,167 @@ explanation_vargrad = craft.compute_explanation_per_concept(
...
+    partial_explainer=sobol,
+    concept_ids=[0, 3, 7],
+)
+
+craft.display_images_per_concept(
+    display_images,
+    concept_maps=sobol_maps,
+    order=[0, 3, 7],
+)
+```
```
AntoninPoche: I do not think a second full example is necessary. I would rather write the API. Then examples only with the part that differ (PartialExplainer).

Also, Rise and Sobol are really close, does it work with gradient-based methods? Maybe Sobol and IntegratedGradients would make better examples.

docs/api/concepts/holistic_craft.md
```diff
@@ -162,14 +195,167 @@ explanation_vargrad = craft.compute_explanation_per_concept(
...
+
+`nb_design` must be a nonzero power of two. Keep Sobol's default `nb_channels=1`: concept
+localization computes a separate single-channel attribution map for each selected concept.
+This differs from Sobol-based concept importance, which attributes the final task prediction
+to concept coefficients.
+
+### Coefficient Maps and Localization Maps
+
+Use `coeffs_u` to display latent coefficient maps:
```
AntoninPoche: Make sure to really define "coefficient map", I would harmonize the term between coefficients and activations.

docs/api/concepts/holistic_craft.md
AntoninPoche, line 1: There is a slop smell to it. notably, with the diversity of terms used to refer to the same thing.

xplique/concepts/holistic_craft.py
```diff
@@ -107,6 +108,129 @@ def __call__(self, model, batch_size):
         return self.explainer_class(model=model, batch_size=batch_size, **self.kwargs)
 
 
+class ConceptLocalizer:
+    """Map input samples to one scalar score per learned concept.
```
AntoninPoche: Insufficient description.

xplique/concepts/holistic_craft.py
```diff
@@ -107,6 +108,129 @@ def __call__(self, model, batch_size):
         return self.explainer_class(model=model, batch_size=batch_size, **self.kwargs)
 
 
+class ConceptLocalizer:
```
AntoninPoche: If the class is not user exposed, it be written `_ConceptLocalizer`. (Not important).

xplique/concepts/holistic_craft.py
```diff
@@ -107,6 +108,129 @@ def __call__(self, model, batch_size):
...
+        ``(batch_size, *spatial_dimensions, number_of_concepts)`` and must return
+        an array with shape ``(batch_size, number_of_concepts)``. Coefficients are
+        reduced as-is, so signed factorizers retain their sign; use a callable
+        reducer such as a mean absolute value when magnitude is desired.
+
+    Returns
+    -------
+    concept_scores
+        Float32 concept scores with shape
+        ``(batch_size, number_of_concepts)``.
```
AntoninPoche: Its weird to have `Returns` for a class.

xplique/concepts/holistic_craft.py
```diff
@@ -107,6 +108,129 @@ def __call__(self, model, batch_size):
...
+            raise RuntimeError(
+                "The selected factorizer cannot encode unseen activations. "
+                "Input-to-concept localization requires factorizer.encode() "
+                "because attribution methods evaluate perturbed inputs."
+            ) from error
+        return self._reduce_coefficients(coeffs_u)
+
+    def __call__(self, inputs: Any) -> np.ndarray:
+        """Return reduced concept scores for a batch of inputs."""
+        return self._compute_scores(inputs)
```
AntoninPoche, line 231: Do you really need both?

Overall, the class have about twice the number of code lines it should require.

xplique/concepts/holistic_craft.py
```diff
@@ -107,6 +108,129 @@ def __call__(self, model, batch_size):
...
+            scores = coeffs_u
+        elif isinstance(self.concept_reducer, str):
+            reducers = {
+                "mean": np.mean,
+                "sum": np.sum,
+                "max": np.max,
+            }
+            scores = reducers[self.concept_reducer](coeffs_u, axis=spatial_axes)
+        else:
+            scores = self.concept_reducer(coeffs_u)
```
AntoninPoche: Storing the correct reducer should be done at init.

xplique/concepts/holistic_craft.py
```diff
@@ -125,8 +249,14 @@ class HolisticCraft(ABC):
     The workflow involves:
     1. Extracting latent activations from a computer vision model
     2. Factorizing activations into interpretable concepts using NMF
-    3. Computing concept importance using gradient-based attribution methods
-    4. Visualizing concepts as spatial heatmaps overlaid on input images
+    3. Computing concept importance by attributing task predictions to concepts
+    4. Visualizing coefficient activation maps overlaid on input images
```
AntoninPoche: Now we have coefficient activation maps, please converge to a single term.

xplique/concepts/holistic_craft.py
```diff
@@ -125,8 +249,14 @@ class HolisticCraft(ABC):
...
+    5. Optionally localizing concept scores to input regions with black-box attribution
```
AntoninPoche: I would say that this is not exactly optional. Rather, we might choose between 4. and 5.

To be clearer, they are both ways to interpret concepts. I would write it has a step 4. Interpret concepts, with two solutions.

Also, do not hesitate to remind the functions and link to their documentation.

xplique/concepts/holistic_craft.py
```diff
@@ -125,8 +249,14 @@ class HolisticCraft(ABC):
...
+
+    Concept activation, concept importance, and concept localization answer different
+    questions. Coefficients describe how strongly a concept is present, importance
+    describes how much it contributes to a task prediction, and localization describes
+    which input regions drive a selected concept score.
```
AntoninPoche: You can put the doc table.

xplique/concepts/holistic_craft.py
```diff
@@ -649,6 +779,298 @@ def make_concept_decoder(self, latent_data: LatentData) -> Any:
...
+        localizer
+            Callable suitable for Xplique black-box explainers, returning
+            float32 scores with shape ``(batch_size, number_of_concepts)``.
+
+        Raises
+        ------
+        ValueError
+            If ``concept_reducer`` is invalid.
+        """
+        return ConceptLocalizer(self, concept_reducer)
```
AntoninPoche: "Sarcasm": What a useful function.

More concretely, when you want to use a different class depending if you are using tf or torch. You add an attribute `concept_localizer_class: type[ConceptLocalizer]`. You set this parameter in both the tf and torch versions.

Then, in the code, you replace `self.make_concept_localizer(...)` by `self.concept_localizer_class(...)`.

If you are lazy, you can even name the attribute `make_concept_localizer` but I would not recommend it.

xplique/concepts/torch/holistic_craft.py
```diff
@@ -190,6 +190,48 @@ def make_concept_decoder(self, latent_data: LatentData) -> TorchWrapper:
...
+
+class ConceptLocalizerTorch(nn.Module, ConceptLocalizer):
+    """PyTorch concept localizer module."""
+
+    def __init__(
+        self,
+        parent_craft: HolisticCraft,
+        concept_reducer: Union[str, Callable] = "mean",
+    ) -> None:
+        super().__init__()
```
AntoninPoche: Do you really need it to be a `nn.Module`?

xplique/concepts/holistic_craft.py
AntoninPoche: Okay, it is even worse then I though, the difference between the inherited torch and tf versions are post-process conversion to tf or numpy. The inheritance process seems completely disproportionate.

I imagine that there are many time where the common part ends up with numpy and you later convert to torch or tf.

Maybe harmonize this conversion and call it anytime you need. That would simplify the whole thing.

Well, I checked, and you pass the localizer to the PartialExplainer, so it needs to be self-contained. But still, that seems more complex than it needs to be.

xplique/concepts/holistic_craft.py
```diff
@@ -649,6 +779,298 @@ def make_concept_decoder(self, latent_data: LatentData) -> Any:
...
+        return images_np.astype(np.float32, copy=False)
+
+    def compute_concept_attributions(
+        self,
+        images,
+        partial_explainer: PartialExplainer,
+        concept_ids: Optional[List[int]] = None,
+        concept_reducer: Union[str, Callable] = "mean",
+    ) -> np.ndarray:
+        """Compute input-space localization maps for selected learned concepts.
```
AntoninPoche, line 891: Harmonize terms.

docs/api/concepts/holistic_craft.md
```diff
@@ -162,14 +195,167 @@ explanation_vargrad = craft.compute_explanation_per_concept(
...
+`(N, C, H, W)` images to the same high-level method; Xplique handles the layout conversion
+for the wrapped localizer.
+
+### Localizing Concepts with RISE
+
+```python
+from xplique.attributions import Rise
+from xplique.concepts import PartialExplainer
+
+rise = PartialExplainer(
```
AntoninPoche: I do not like the name. But you told me it matches another part of the pipeline.

xplique/concepts/holistic_craft.py
```diff
@@ -649,6 +779,298 @@ def make_concept_decoder(self, latent_data: LatentData) -> Any:
...
+            raise TypeError(
+                f"partial_explainer must be a PartialExplainer instance, got "
+                f"{type(partial_explainer).__name__}."
+            )
+
+        if partial_explainer.kwargs.get("operator") is not None:
+            raise ValueError(
+                "Concept localization uses one-hot concept targets and does not accept a "
+                "custom operator. Omit the operator argument."
+            )
```
AntoninPoche: Could just ignore it. Or raise a warning.

xplique/concepts/holistic_craft.py
```diff
@@ -649,6 +779,298 @@ def make_concept_decoder(self, latent_data: LatentData) -> Any:
...
+        explainer_class = partial_explainer.explainer_class
+        # Class-based white-box explainers can be rejected explicitly. Callable factories
+        # are intentionally left to their own validation because their class is unknown.
+        if isinstance(explainer_class, type) and issubclass(explainer_class, WhiteBoxExplainer):
+            raise ValueError(
+                "Input-to-concept localization currently supports black-box attribution "
+                "methods only. The ConceptLocalizer uses factorizer.encode(), which is "
+                "not guaranteed to be differentiable. Use Rise, SobolAttributionMethod, "
+                "Occlusion, Lime, KernelShap, or another compatible black-box explainer."
+            )
```
AntoninPoche, line 965: I might have missed it in the doc, otherwise, it should be specified.

xplique/concepts/holistic_craft.py
```diff
@@ -649,6 +779,298 @@ def make_concept_decoder(self, latent_data: LatentData) -> Any:
...
+
+        return images_np.astype(np.float32, copy=False)
+
+    def compute_concept_attributions(
+        self,
+        images,
+        partial_explainer: PartialExplainer,
+        concept_ids: Optional[List[int]] = None,
+        concept_reducer: Union[str, Callable] = "mean",
+    ) -> np.ndarray:
```
AntoninPoche: With default parameters, depending on the number of concepts, this might take a really long time (nb concepts x images x nb perturbations / batch size). Maybe put few comments in the doc for this. (I like inline comments in the code examples).

xplique/concepts/holistic_craft.py
```diff
@@ -649,6 +779,298 @@ def make_concept_decoder(self, latent_data: LatentData) -> Any:
...
+            if single_concept_map.shape != expected_shape:
+                raise ValueError(
+                    "Concept attribution map shape must match attribution inputs spatial "
+                    f"shape {expected_shape}, got {single_concept_map.shape}."
+                )
+            if not np.all(np.isfinite(single_concept_map)):
+                raise ValueError(
+                    f"Concept attribution map for concept {concept_id} must contain "
+                    "only finite values."
+                )
```
AntoninPoche: These tests look like unit tests, not inline assertions.

xplique/concepts/holistic_craft.py
```diff
@@ -705,36 +1127,8 @@ def _prepare_display_concept_inputs(
...
-                for concept_id in concepts_id
-            ):
-                raise ValueError(
-                    f"order concept IDs must be integers between 0 and "
-                    f"{self.number_of_concepts - 1}"
-                )
-            if len(set(concepts_id)) != len(concepts_id):
-                raise ValueError("order cannot contain duplicate concept IDs")
+        images_np = self._normalize_image_batch_to_nhwc(images)
+        concepts_id = self._validate_concept_ids(order, parameter_name="order")
```
AntoninPoche: This looks like adding a call layer without justification.

You should replace the `_prepare_display_concept_inputs` by the two above calls.

xplique/concepts/holistic_craft.py
```diff
@@ -823,15 +1236,32 @@ def display_images_per_concept(
...
-        images_np, coeffs_u, concepts_id = self._prepare_display_concept_inputs(
-            images, coeffs_u, order
-        )
+        if concept_maps is None:
+            images_np, coeffs_u, concepts_id = self._prepare_display_concept_inputs(
+                images, coeffs_u, order
+            )
+            heatmap_source = coeffs_u
+        else:
+            images_np = self._normalize_image_batch_to_nhwc(images)
```
AntoninPoche: In both case the `self._normalize_image_batch_to_nhwc(images)` is the first line called and do not depend on the concept_maps.

tests/concepts/test_holistic_craft_classification_torch.py
```diff
@@ -3,22 +3,47 @@
...
+
+    def fit(self, activations):
+        self.is_fitted = True
+        return np.eye(2, dtype=np.float32), np.asarray(activations, dtype=np.float32)
+
+    def encode(self, activations):
+        return np.asarray(activations, dtype=np.float32)
+
+    def encode_differentiable(self, activations):
+        return activations
```
AntoninPoche: Looks like you can make this a pytest fixture rather than copying it between the two test files.

tests/concepts/test_holistic_craft_classification_torch.py
```diff
@@ -356,3 +405,107 @@ def test_craft_sobol_importance(image_data, craft_data):
...
+        concept_ids=[0],
+    )
+
+    assert maps.shape == (1, 4, 4, craft.number_of_concepts)
+    assert maps.dtype == np.float32
+    assert np.all(np.isfinite(maps[..., 0]))
+    assert np.all(np.isnan(np.delete(maps, 0, axis=-1)))
+
+
+@pytest.mark.parametrize("reducer", ["mean", "sum"])
```
AntoninPoche: Not "max"?

AntoninPoche (changes requested):
There are few artifacts due to AI usage:
- Different names for the same thing
- Overcomplex code
- Lack of comments
- Many tests with weird structure, this will surely force future change to follow an arbitraly defined architecture. We called that technical debt the last time. (I did not review in detail the regression one).

Nonetheless, it seems to do the job, depends how much effort you want to put into it.

Agustin-Picard:
I've updated the notebook following your suggestions @AntoninPoche : https://drive.google.com/file/d/19eg2uU95A_LW1hQ1ENrhGI41lLK35UHu/view?usp=sharing

docs/api/concepts/holistic_craft.md
Agustin-Picard: Good idea, done ;)

docs/api/concepts/holistic_craft.md
Agustin-Picard: I've changed it to put it in terms of perturbations, which I think makes it much clearer.

docs/api/concepts/holistic_craft.md
Agustin-Picard: I've clarified notation below (and removed coefficient heatmap, which made sense in terms of code but not in XAI terms)

docs/api/concepts/holistic_craft.md
Agustin-Picard: The diagram doesn't make much sense as it is, but it was supposed to mean: "you choose an input image, you choose an operator for the attribution (a class type xplique operator per concept to explain), with these elements we apply attribution from the input and get the concept localization". I've modified it entirely so it should be much clearer now.

docs/api/concepts/holistic_craft.md
Agustin-Picard: I does work for CNNs as we rearrange('b,h,w,c'->'bhw,c') before learning the feature extractor, so it works in a similar way as ViTs, but does suffer from "limited readability" depending on perceptive field and architecture (ie, heatmaps get a bit noisy).

docs/api/concepts/holistic_craft.md
Agustin-Picard: Changed it ;)

docs/api/concepts/holistic_craft.md
Agustin-Picard: Yup, it was too strict for no reason. Adjusted it, should be better now.

docs/api/concepts/holistic_craft.md
Agustin-Picard: For the moment, it does not work with white box methods (as it requires a factorizer supporting `encode_differentiable`), but should be coming up in a PR soon. I wanted to get the API set up first in this PR.

docs/api/concepts/holistic_craft.md
Agustin-Picard: So I removed it, now it's just concept activation, which is much clearer IMO

docs/api/concepts/holistic_craft.md
Agustin-Picard: Fix it (I think...)

xplique/concepts/holistic_craft.py
Agustin-Picard: Should be much better now.

xplique/concepts/holistic_craft.py
Agustin-Picard: You're right, changed it to make it protected as it didn't need to be public really.

xplique/concepts/holistic_craft.py
Agustin-Picard: It was for the __call__ method ^^

xplique/concepts/holistic_craft.py
Agustin-Picard: Fixed

xplique/concepts/holistic_craft.py
Agustin-Picard: Removed it

xplique/concepts/torch/holistic_craft.py
Agustin-Picard: Actually, not in this version, so I removed it

docs/api/concepts/holistic_craft.md
Agustin-Picard: Yup, comes from the HolisticCraft PR, and is used for the concept importance part, so it made sense to use the same here.

xplique/concepts/holistic_craft.py
Agustin-Picard: Fixed it.

xplique/concepts/holistic_craft.py
Agustin-Picard: It's in the documentation and the tutorial normally

xplique/concepts/holistic_craft.py
Agustin-Picard: It's normally clarified in the doc and the tutorial that this is a pretty expensive operation (compared to the CAM)

xplique/concepts/holistic_craft.py
Agustin-Picard: I simplified it a bit, but I still believe Xplique doesn't catch unsanitized inputs enough in general, leading to some mysterious errors sometimes.

tests/concepts/test_holistic_craft_classification_torch.py
Agustin-Picard: Place it in a conftest file for the concepts. We could move more general stuff over there later, but it's outside the scope of this PR.

tests/concepts/test_holistic_craft_classification_torch.py
Agustin-Picard: Added it

xplique/concepts/holistic_craft.py
Agustin-Picard: You're right, working on a simplification ;)

Agustin-Picard:
Thanks again @AntoninPoche ! Since the last review, I pushed a broader cleanup/simplification pass addressing most of the points you raised:

- Harmonized the terminology and clarified the distinction between **concept importance** (perturb/attribute concept activations → task prediction) and **concept localization** (perturb input → concept activation score). Removed all mentions of coefficient maps, as concept activation maps is much clearer.
- Simplified the implementation further (as you suggested): there is now a single shared concept localizer in `HolisticCraft` for TF and torch; and the extra Torch wrapper layer was removed. The Torch backend now only handles the NHWC → NCHW conversion through `_prepare_localizer_inputs` (and is basically a no-op for TF).
- Simplified the display/localization plumbing and removed unnecessary test scaffolding.
- Reorganized the tests into dedicated localization/parity coverage, removed duplicated TF/Torch localizer tests, and kept parity checks for `mean` / `sum` / `max` (including CUDA when available).
- Updated the documentation and notebook accordingly (keeping mostly what you had proposed, with maybe 2 or 3 modifications).

I've also adjusted the PR's description to reflect the current API!

docs/api/concepts/holistic_craft.md
```diff
@@ -53,21 +53,45 @@ The builder takes the model and a layer index to define the split point. Everyth
...
+# 1. Fit concepts.
+craft.fit(images)
+# 2. Compute concept activation maps.
+coeffs_u = craft.transform(images)
+# 3. Estimate concept importance.
+importances = craft.estimate_importance(images, operator, class_id)
+# 4. Interpret concept activation maps.
+craft.display_images_per_concept(images, coeffs_u=coeffs_u)
+# Alternatively, compute concept localization maps with a PartialExplainer.
+concept_maps = craft.attribute_concepts_to_inputs(images, partial_explainer)
```
AntoninPoche: Mismatch between steps

xplique/concepts/holistic_craft.py
AntoninPoche: I still do not understand the need for this method.

`localizer = self.make_concept_localizer(concept_reducer)` -> `localizer = self._ConceptLocalizer(self, concept_reducer)`

Would result in the removal of the `make_concept_localizer` method.

AntoninPoche (approved):
LGTM!

Much clearer on the documentation and efficient in the code!

PR #187, opened 2026-09-21
Stabilize PyTorch CRAFT shape test

Agustin-Picard:
Stabilize the PyTorch CRAFT shape test that could fail intermittently during NMF
importance estimation.

## Root Cause

The test generated random data and model weights. A random convolution followed by
ReLU could produce all-zero activations, causing NMF to learn an all-zero concept
bank. Newer scikit-learn validation then rejected that bank during `transform()`.

## Changes

- Generate deterministic inputs and labels.
- Initialize positive, distinct convolution weights and bias.
- Run the test model in evaluation mode to disable dropout.
- Assert that activations and the fitted concept bank are non-degenerate.


AntoninPoche (approved):
LGTM

